PR #3, opened 2025-05-09
Handle url sanitization

shikharbsar:


src/oauth-manager/oauth-utils.ts
```diff
@@ -498,6 +498,41 @@ export interface ParsedApprovalResult {
...
+        const urlWithProtocol = trimmedUrl.startsWith('http://') || trimmedUrl.startsWith('https://') 
+            ? trimmedUrl 
+            : `https://${trimmedUrl}`;
+        
+        const parsedUrl = new URL(urlWithProtocol);
+        
+        // Ensure it's using HTTPS
+        if (parsedUrl.protocol !== 'https:') {
+            throw new Error('URL must use HTTPS');
+        }
```
ashubham, line 522: On line 513 you allowed `http` and now you are asserting `https` only. We should allow `http` as well in my opinion as nebula dev clusters would be `http`.

shikharbsar: My bad, removed this

ashubham (approved):

PR #54, opened 2025-07-29
Png image

shikharbsar:


src/servers/mcp-server.ts
```diff
@@ -224,6 +224,21 @@ export class MCPServer extends BaseMCPServer {
...
+            console.log("[DEBUG] Storing token in KV");
+            const token = crypto.randomUUID();
+            const tokenData = {
+                sessionId: answer.session_identifier,
+                generationNo: answer.generation_number,
+                instanceURL: this.ctx.props.instanceUrl,
+                accessToken: this.ctx.props.accessToken
+            };
+            await putInKV(token, tokenData, this.ctx.env);
+            tokenUrl = `${this.ctx.env?.HOST_NAME}/data/img?token=${token}`;
```
ashubham, line 240: Maybe we call `token` something like `uniqueId` or `tempId` to not make the agent believe its a token protected resource and should not be opened

src/servers/openai-mcp-server.ts
```diff
@@ -131,15 +132,44 @@ export class OpenAIDeepResearchMCPServer extends BaseMCPServer {
...
+            console.log("[DEBUG] Storing token in KV");
+            const token = crypto.randomUUID();
+            const tokenData = {
+                sessionId: answer.session_identifier,
+                generationNo: answer.generation_number,
+                instanceURL: this.ctx.props.instanceUrl,
+                accessToken: this.ctx.props.accessToken
+            };
+            await putInKV(token, tokenData, this.ctx.env);
+            tokenUrl = `${this.ctx.env?.HOST_NAME}/data/img?token=${token}`;
```
ashubham, line 156: this seems to be duplicate code

src/servers/openai-mcp-server.ts
```diff
@@ -131,15 +132,44 @@ export class OpenAIDeepResearchMCPServer extends BaseMCPServer {
...
+        }
+        let content = answer.data;
+        if (tokenUrl) {
```
ashubham, line 159: will tokenUrl be not available in any case ? If not we can remove this conditionality

src/servers/openai-mcp-server.ts
```diff
@@ -131,15 +132,44 @@ export class OpenAIDeepResearchMCPServer extends BaseMCPServer {
...
+                content = `Data: ${answer.data}
+
+**CHART VISUALIZATION AVAILABLE:**
+There is a data visualization chart for this query available at: "${tokenUrl}".
+
+This URL contains a PNG image showing the visual representation of the above data for the question: "${question}". Please fetch and display this image to provide a complete analysis with both the data and its visual representation. The image must be displayed in the response of this question. 
+Use just this image in the response, do not use any other images or visualizations or charts. It may take time to get the image, so please wait for the image to be fetched and displayed in the response.`;
```
ashubham, line 166: Can we make this consistent across different mcp servers ? Maybe move to the common service layer.

wrangler.jsonc
```diff
@@ -44,7 +44,8 @@
     ],
 	"vars": {
 		"HONEYCOMB_DATASET": "",
-		"HONEYCOMB_API_KEY": ""
+		"HONEYCOMB_API_KEY": "",
+		"HOST_NAME": "https://agent.thoughtspot.app"
```
ashubham, line 48: We have created aliases for this, and might also change this in the future. Lets not hardcode this. This should be available in the request payload headers.

ashubham (changes requested):

src/servers/mcp-server.ts
shikharbsar: Done

src/servers/openai-mcp-server.ts
shikharbsar: done

src/servers/openai-mcp-server.ts
shikharbsar: done

wrangler.jsonc
shikharbsar: Yeah, added it as part of props, let me know if I need to change it.

src/handlers.ts
```diff
@@ -174,13 +176,45 @@ class Handler {
...
         span?.setStatus({ code: SpanStatusCode.OK, message: "Token stored successfully" });
 
         return { redirectTo };
     }
+
+    @WithSpan('get-answer-image')
+    async getAnswerImage(request: Request, env: Env) {
+        const span = getActiveSpan();
+        const url = new URL(request.url);
+        const token = url.searchParams.get('token') || url.searchParams.get('uniqueId') || '';
```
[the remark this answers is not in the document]
ashubham: Rename the variable as well

src/servers/mcp-server-base.ts
```diff
@@ -147,6 +148,41 @@ export abstract class BaseMCPServer extends Server {
         this.addTracker(mixpanel);
     }
 
+    protected async getTokenUrl(sessionId: string, generationNo: number) {
+        let tokenUrl = "";
+        // Generate token and store in KV store
+        if (this.ctx.env?.OAUTH_KV) {
```
ashubham, line 154: Lets put this in a `try ... catch`. The `putInKV` method should throw if `OAUTH_KV` is not available. And we should handle that condition in the `catch`

src/servers/mcp-server-base.ts
```diff
@@ -147,6 +148,41 @@ export abstract class BaseMCPServer extends Server {
...
+            const token = crypto.randomUUID();
+            const tokenData = {
+                sessionId: sessionId,
+                generationNo: generationNo,
+                instanceURL: this.ctx.props.instanceUrl,
+                accessToken: this.ctx.props.accessToken
+            };
+            await putInKV(token, tokenData, this.ctx.env);
+            const hostName = this.ctx.props.hostName;
+            if (hostName && (hostName.startsWith("http") || hostName.startsWith("https"))) {
```
ashubham, line 165: handle this while setting the `prop`

src/thoughtspot/thoughtspot-service.ts
```diff
@@ -283,6 +283,18 @@ export class ThoughtSpotService {
         return liveboardUrl;
     }
 
+    @WithSpan('get-answer-image')
+    async getAnswerImagePNG(sessionId: string, GenNo: number): Promise<HttpFile> {
```
ashubham, line 287: genNo

src/utils.ts
```diff
@@ -92,20 +93,75 @@ export class McpServerError extends Error {
...
+        get server(): T {
+            if (!this._server) {
+                const context: Context = {
+                    props: this.props, // Available after McpAgent._init() sets it
+                    env: this._env     // Stored from constructor
+                };
+                this._server = new MCPServer(context);
+            }
+            return this._server;
+        }
```
ashubham, line 127: declare public env: Env;

And then move creation of the server inside the `init` method.

PR #57, opened 2025-07-31
add open api spec for tools

shikharbsar:


public/docs/glean-actions.txt
```diff
@@ -0,0 +1,19 @@
+- Glean actions
```
ashubham, line 1: Make this .md ?

ashubham (approved):

PR #59, opened 2025-08-06
Datasource discovery tool

shikharbsar:


src/servers/mcp-server.ts
```diff
@@ -120,7 +126,17 @@ export class MCPServer extends BaseMCPServer {
                         readOnlyHint: true,
                         destructiveHint: false,
                     },
-                }
+                },
+                ...(this.isVersionGreaterThan1013() ? [{
```
ashubham, line 130: put comment on why the version check is needed

ashubham: Maybe instead of the version check, lets move to using the flag for this feature perhaps.

src/servers/openai-mcp-server.ts
```diff
@@ -99,10 +99,21 @@ export class OpenAIDeepResearchMCPServer extends BaseMCPServer {
...
+            if (!dataSources || dataSources.length === 0) {
+                return this.createErrorResponse("No data source suggestions found", "No data source suggestions found");
+            }
+            // Use the first (highest confidence) data source for getting relevant questions
+            const primaryDataSource = dataSources[0];
+            datasourceIdToUse = primaryDataSource.header.guid;
+        }
+        // If the datasource id is present, get the relevant questions for the query
+        if (datasourceIdToUse) {
+            const relevantQuestions = await this.getThoughtSpotService().getRelevantQuestions(queryWithoutDatasourceId, [datasourceIdToUse], "");
```
ashubham, line 116: We want to let the agent choose the datasource for us. We should return the list of sources as search results. And then the agent will do a search again with the relevant datasource added to the query.

src/thoughtspot/thoughtspot-service.ts
```diff
@@ -10,6 +10,73 @@ import type { DataSource, SessionInfo } from "./types";
...
+                span?.setAttribute("suggestions_count", 0);
+                return null;
+            }
+            
+            span?.setAttribute("suggestions_count", response.dataSources.length);
+            
+            // Sort data sources by confidence in descending order
+            const sortedDataSources = [...response.dataSources].sort((a, b) => b.confidence - a.confidence);
+            
+            const highestConfidenceDataSource = sortedDataSources[0];
```
ashubham, line 55: do we really need to sort this again ?

src/servers/openai-mcp-server.ts
shikharbsar: done

PR #119, opened 2026-04-23
SCAL-308277 Add metrics recorder core

kanwarbajwa:
## Summary

Introduce the request-scoped metrics runtime core for the Worker and add focused tests so the new runtime metrics files do not regress repo coverage.

## What Changed

- add metric definitions, label normalization, and runtime config for sink selection
- add the request-scoped recorder, sink interfaces, composite/noop sinks, and context helpers
- wrap the Worker fetch entrypoint so every request gets a recorder and non-blocking flush behavior
- add targeted tests for metric context, sink behavior, recorder branches, runtime config, and request-scoped recorder helpers

## Notes

- this PR only adds the runtime core, request wrapper plumbing, and tests
- external Analytics Engine and Grafana emission lands in follow-up PRs

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -11,6 +11,7 @@ import { instrumentedMCPServer } from "./cloudflare-utils";
 import { MCPServer } from "./servers/mcp-server";
 import { apiServer } from "./servers/api-server";
 import { withBearerHandler } from "./bearer";
+import { withRequestMetrics } from "./metrics/runtime/request-metrics";
 import { OpenAIDeepResearchMCPServer } from "./servers/openai-mcp-server";
 import { ConversationStorageServer } from "./servers/conversation-storage-server";
 
@@ -118,11 +119,13 @@ export default {
 		env: Env,
 		ctx: ExecutionContext,
 	): Promise<Response> {
-		if (HEADERS_TO_STRIP.some((header) => request.headers.has(header))) {
-			const headers = new Headers(request.headers);
-			HEADERS_TO_STRIP.forEach((header) => headers.delete(header));
-			request = new Request(request, { headers });
-		}
-		return instrumentedOAuthHandler.fetch!(request, env, ctx);
+		return withRequestMetrics(env as unknown as Record<string, unknown>, ctx, async () => {
+			if (HEADERS_TO_STRIP.some((header) => request.headers.has(header))) {
+				const headers = new Headers(request.headers);
+				HEADERS_TO_STRIP.forEach((header) => headers.delete(header));
+				request = new Request(request, { headers });
+			}
+			return instrumentedOAuthHandler.fetch!(request, env, ctx);
+		});
 	},
 };
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,17 +1,17 @@
-import { trace } from "@opentelemetry/api";
+import OAuthProvider from "@cloudflare/workers-oauth-provider";
 import {
-	instrument,
 	type ResolveConfigFn,
 	type TraceConfig,
+	instrument,
 } from "@microlabs/otel-cf-workers";
-import OAuthProvider from "@cloudflare/workers-oauth-provider";
+import { trace } from "@opentelemetry/api";
 
-import handler from "./handlers";
-import { instrumentedMCPServer } from "./cloudflare-utils";
-import { MCPServer } from "./servers/mcp-server";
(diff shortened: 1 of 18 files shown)


kanwarbajwa:
@claude please review

src/metrics/runtime/metrics-recorder.ts
```diff
@@ -0,0 +1,111 @@
...
+
+			await this.options.sink.flush({
+				observations: this.snapshot(),
+				resourceAttributes: { ...this.options.resourceAttributes },
+			});
+		} catch (error) {
+			console.error("[metrics] Flush failed", error);
+		} finally {
+			this.flushed = true;
+		}
```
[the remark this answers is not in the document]
kanwarbajwa: flushInternal() now closes the recorder before awaiting the sink, so metrics recorded during an in-flight flush are rejected instead of getting stranded. I also added a race-specific test to cover that case.

kanwarbajwa:
@copilot review this PR

kanwarbajwa:
@gemini-code-assist review again

src/metrics/runtime/metric-context.ts
```diff
@@ -0,0 +1,133 @@
+import type {
+	ApiSurface,
+	AuthMode,
+	RouteGroup,
+	StatusClass,
+	Transport,
+} from "./metric-types";
+
+export function getRouteGroup(pathname: string): RouteGroup {
+	switch (pathname) {
```
Rifdhan: How will we ensure this and all the other complex path-splitting logic below get updated if a new path is added in the future?

src/index.ts
```diff
@@ -115,11 +116,13 @@ export default {
...
 		ctx: ExecutionContext,
 	): Promise<Response> {
-		if (HEADERS_TO_STRIP.some((header) => request.headers.has(header))) {
-			const headers = new Headers(request.headers);
-			HEADERS_TO_STRIP.forEach((header) => headers.delete(header));
-			request = new Request(request, { headers });
-		}
-		return instrumentedOAuthHandler.fetch!(request, env, ctx);
+		return withRequestMetrics(env as unknown as Record<string, unknown>, ctx, async () => {
+			if (HEADERS_TO_STRIP.some((header) => request.headers.has(header))) {
```
Rifdhan: My intention here was to strip the headers BEFORE metrics gets involved, because I want to avoid any future problem where a tracing header messes up the metrics init like we just had. Should we rearrange this accordingly?

package-lock.json
```diff
@@ -48,58 +48,6 @@
         "express": "^5.1.0"
       }
     },
-    "node_modules/@ai-sdk/gateway": {
```
Rifdhan, line 51: Why are there changes to package-lock? This PR is not changing package.json so it should not touch this either

kanwarbajwa: I'm not sure how this commit got here. The commit says: Copilot and kanwarbaja authored this. I'm not sure what this is. I'm removing this commit.

src/index.ts
kanwarbajwa: Fixed. I moved tracing-header stripping ahead of withRequestMetrics(...) so the request is sanitized before any metrics or downstream instrumentation path runs.

src/metrics/runtime/metric-context.ts
kanwarbajwa: Fixed. Previously the route classification was spread across multiple helpers, so a new public path could require updates in several places. I consolidated it into a single shared route-context table plus one resolver.

Rifdhan: What happens if we add a new route but there is no config in this file? I feel it might be good to throw on such cases so that devs are forced to update it here. If we don't have a good mechanism to get eyes on it, this can become stale and not get updated.

kanwarbajwa: I agree we should force updates when new public routes are added, but I would avoid throwing on the runtime request path because metrics should not be able to break business logic. I moved the known public routes into  constants and added a test guard that verifies every exact public route requiring metrics has an explicit metrics context. This gives us CI enforcement without making request handling depend on metrics config correctness.

Rifdhan: Sounds good

Rifdhan (approved):

ashubham (approved):

Rifdhan (approved):

PR #125, opened 2026-04-29
SCAL-308278 Add Analytics Engine metrics sink

kanwarbajwa:
## Summary

Add the Analytics Engine sink for the metrics runtime so PM and tenant event observations can be written to the existing Cloudflare `ANALYTICS` binding.

## What Changed

- add an Analytics Engine sink that writes one datapoint per metric observation
- define the positional Analytics Engine schema for indexes, blobs, and doubles
- classify metric names into stable event families for PM-oriented querying
- wire request metrics to use `env.ANALYTICS` when the analytics sink is enabled
- keep missing bindings as noop through the existing sink selection path
- cover datapoint mapping, binding detection, and default sink wiring in tests

## Semantics

The schema reserves index positions for tenant and user identity. This PR wires the sink and raw event shape; later instrumentation PRs will populate those identity fields from ThoughtSpot session context where available.

### DIFF
--- a/src/metrics/runtime/analytics-engine-sink.ts
+++ b/src/metrics/runtime/analytics-engine-sink.ts
@@ -1,0 +1,197 @@
+import {
+	METRIC_NAMES,
+	type MetricLabels,
+	type MetricName,
+} from "./metric-types";
+import type {
+	MetricObservation,
+	MetricResourceAttributes,
+	MetricsFlushPayload,
+	MetricsSink,
+} from "./metrics-sink";
+
+export type AnalyticsEngineDataPointLike = {
+	indexes?: ((ArrayBuffer | string) | null)[];
+	blobs?: ((ArrayBuffer | string) | null)[];
+	doubles?: number[];
+};
+
+export type AnalyticsEngineDatasetLike = {
+	writeDataPoint(event?: AnalyticsEngineDataPointLike): void;
+};
+
+export const ANALYTICS_ENGINE_SCHEMA_VERSION = "mcp_metrics_v1";
+
+export const ANALYTICS_ENGINE_INDEX_FIELDS = [
+	"schema_version",
+	"event_family",
+	"metric_name",
+	"tenant_id",
+	"user_id",
+] as const;
+
+export const ANALYTICS_ENGINE_BLOB_FIELDS = [
+	"metric_kind",
+	"route_group",
+	"transport",
+	"auth_mode",
+	"api_surface",
+	"api_version",
+	"outcome",
+	"status_class",
+	"tool_name",
+	"upstream_operation",
+	"message_type",
+	"is_thinking",
+	"is_done",
+	"operation",
+	"deployment_environment",
+	"service_name",
+	"service_namespace",
+	"service_version",
+] as const;
+
+export const ANALYTICS_ENGINE_DOUBLE_FIELDS = [
+	"metric_value",
+	"timestamp_ms",
+] as const;
+
+type AnalyticsEngineIdentity = {
+	tenantId?: string;
+	userId?: string;
+};
+
+type AnalyticsEngineMetricFamily =
+	| "analysis"
+	| "auth"
+	| "dashboard"
+	| "http"
+	| "resource"
+	| "stream_storage"
+	| "tool"
+	| "upstream";
+
+function nullableString(value: string | undefined): string | null {
+	return value && value.length > 0 ? value : null;
+}
+
+function getLabel(
+	labels: MetricLabels,
+	key: keyof MetricLabels,
+): string | null {
+	return nullableString(labels[key]);
+}
+
+function getResourceAttribute(
+	resourceAttributes: MetricResourceAttributes,
+	key: keyof MetricResourceAttributes,
+): string | null {
+	return nullableString(resourceAttributes[key]);
+}
+
+export function getAnalyticsEngineMetricFamily(
+	name: MetricName,
+): AnalyticsEngineMetricFamily {
+	switch (name) {
+		case METRIC_NAMES.httpRequestsTotal:
+		case METRIC_NAMES.httpRequestDurationMs:
+		case METRIC_NAMES.httpInflightRequests:
+			return "http";
+		case METRIC_NAMES.toolCallsTotal:
+		case METRIC_NAMES.toolDurationMs:
+			return "tool";
+		case METRIC_NAMES.resourceReadsTotal:
+			return "resource";
+		case METRIC_NAMES.oauthAuthorizeRequestsTotal:
+		case METRIC_NAMES.oauthAuthorizeSubmitTotal:
+		case METRIC_NAMES.oauthCallbackTotal:
+		case METRIC_NAMES.oauthStoreTokenTotal:
+		case METRIC_NAMES.bearerAuthRequestsTotal:
+			return "auth";
+		case METRIC_NAMES.upstreamCallsTotal:
+		case METRIC_NAMES.upstreamDurationMs:
+		case METRIC_NAMES.upstreamStreamsStartedTotal:
+		case METRIC_NAMES.upstreamStreamMessagesTotal:
+			return "upstream";
+		case METRIC_NAMES.analysisSessionsCreatedTotal:
+		case METRIC_NAMES.analysisMessagesSentTotal:
+		case METRIC_NAMES.analysisUpdatesPolledTotal:
+		case METRIC_NAMES.analysisPollWaitMs:
+		case METRIC_NAMES.analysisFirstBufferedUpdateMs:
+		case METRIC_NAMES.analysisFirstPollDelayMs:
+		case METRIC_NAMES.analysisFirstNonEmptyResponseMs:
+		case METRIC_NAMES.analysisSessionsNeverPolledTotal:
+			return "analysis";
+		case METRIC_NAMES.streamStorageErrorsTotal:
+			return "stream_storage";
+		case METRIC_NAMES.dashboardsCreatedTotal:
+		case METRIC_NAMES.dashboardTilesCount:
+			return "dashboard";
+	}
+}
+
+export function toAnalyticsEngineDataPoint(
+	observation: MetricObservation,
+	resourceAttributes: MetricResourceAttributes,
+	identity: AnalyticsEngineIdentity = {},
+): AnalyticsEngineDataPointLike {
+	return {
+		indexes: [
+			ANALYTICS_ENGINE_SCHEMA_VERSION,
+			getAnalyticsEngineMetricFamily(observation.name),
+			observation.name,
+			nullableString(identity.tenantId),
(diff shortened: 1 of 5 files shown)


src/metrics/runtime/analytics-engine-sink.ts
```diff
@@ -0,0 +1,197 @@
...
+		case METRIC_NAMES.analysisFirstNonEmptyResponseMs:
+		case METRIC_NAMES.analysisSessionsNeverPolledTotal:
+			return "analysis";
+		case METRIC_NAMES.streamStorageErrorsTotal:
+			return "stream_storage";
+		case METRIC_NAMES.dashboardsCreatedTotal:
+		case METRIC_NAMES.dashboardTilesCount:
+			return "dashboard";
+	}
+}
```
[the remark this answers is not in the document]
kanwarbajwa: Added `sessionsStartedTotal` to the `auth` family and made the switch exhaustive with a `never` check plus an explicit throw, so future `MetricName` additions cannot silently return `undefined`

test/metrics/runtime/analytics-engine-sink.spec.ts
```diff
@@ -0,0 +1,119 @@
...
+		).toBe("upstream");
+		expect(
+			getAnalyticsEngineMetricFamily(
+				METRIC_NAMES.analysisFirstNonEmptyResponseMs,
+			),
+		).toBe("analysis");
+		expect(
+			getAnalyticsEngineMetricFamily(METRIC_NAMES.dashboardTilesCount),
+		).toBe("dashboard");
+	});
```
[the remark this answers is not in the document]
kanwarbajwa: The test now iterates over all `Object.values(METRIC_NAMES)` and verifies every metric maps to a valid Analytics Engine event family. Also kept an explicit assertion for `sessionsStartedTotal`.

kanwarbajwa:
@gemini-code-assist please review again

src/metrics/runtime/analytics-engine-sink.ts
```diff
@@ -0,0 +1,204 @@
...
+
+export class AnalyticsEngineMetricsSink implements MetricsSink {
+	constructor(private readonly dataset: AnalyticsEngineDatasetLike) {}
+
+	async flush(payload: MetricsFlushPayload): Promise<void> {
+		for (const observation of payload.observations) {
+			this.dataset.writeDataPoint(
+				toAnalyticsEngineDataPoint(observation, payload.resourceAttributes),
+			);
+		}
```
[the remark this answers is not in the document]
kanwarbajwa: Each Analytics Engine data point write is now awaited and isolated with a per-observation `try/catch`, so one failed write does not abort the rest of the flush. I did not add sink-level `isFlushing`; the recorder already owns the flush race guard by setting `flushed = true` before snapshotting and before any await.

kanwarbajwa:
@gemini-code-assist please review again

src/metrics/runtime/analytics-engine-sink.ts
```diff
@@ -0,0 +1,211 @@
...
+				await this.dataset.writeDataPoint(
+					toAnalyticsEngineDataPoint(observation, payload.resourceAttributes),
+				);
+			} catch (error) {
+				console.warn(
+					`[metrics] Failed to write Analytics Engine data point for ${observation.name}`,
+					error,
+				);
+			}
+		}
```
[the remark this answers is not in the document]
kanwarbajwa: Analytics Engine writes now run via `Promise.all` while preserving per-observation error isolation. I intentionally did not add sink-level `isFlushing` because this sink does not accept new metrics. The request recorder owns that race guard and already sets `flushed = true` before snapshotting and before any await.

kanwarbajwa:
@gemini-code-assist please review again

src/metrics/runtime/analytics-engine-sink.ts
```diff
@@ -0,0 +1,213 @@
...
+
+export class AnalyticsEngineMetricsSink implements MetricsSink {
+	constructor(private readonly dataset: AnalyticsEngineDatasetLike) {}
+
+	async flush(payload: MetricsFlushPayload): Promise<void> {
+		await Promise.all(
+			payload.observations.map(async (observation) => {
+				try {
+					await this.dataset.writeDataPoint(
+						toAnalyticsEngineDataPoint(observation, payload.resourceAttributes),
```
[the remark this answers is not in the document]
kanwarbajwa: Agree. Identity should not be inferred from metric labels/observations because those are intentionally low-cardinality and shared with Grafana. I think the right contract is an optional request/event identity field on `MetricsFlushPayload`, which Analytics Engine can use and Grafana can ignore. Current PR can leave it unset, but this makes the future tenant/user enrichment path explicit.

kanwarbajwa:
@gemini-code-assist please review again

src/metrics/runtime/analytics-engine-sink.ts
```diff
@@ -0,0 +1,217 @@
...
+						),
+					);
+				} catch (error) {
+					console.warn(
+						`[metrics] Failed to write Analytics Engine data point for ${observation.name}`,
+						error,
+					);
+				}
+			}),
+		);
```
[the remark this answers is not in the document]
kanwarbajwa: For Cloudflare Analytics Engine, `writeDataPoint` is typed as synchronous `void`, so `Promise.all` adds promise/microtask overhead without real concurrency. I’ll switch this back to a simple `for...of` with per-observation `try/catch`. I’m still not adding sink-level `isFlushing`; `RequestMetricsRecorder` owns metric acceptance and already sets `flushed = true` before snapshotting and before any await.

src/metrics/runtime/analytics-engine-sink.ts
```diff
@@ -0,0 +1,215 @@
...
+	doubles?: number[];
+};
+
+export type AnalyticsEngineDatasetLike = {
+	writeDataPoint(event?: AnalyticsEngineDataPointLike): void;
+};
+
+export const ANALYTICS_ENGINE_SCHEMA_VERSION = "mcp_metrics_v1";
+
+export const ANALYTICS_ENGINE_INDEX_FIELDS = [
```
Rifdhan, line 26: Should we have conversation_id somewhere here or in the next list?

src/metrics/runtime/analytics-engine-sink.ts
```diff
@@ -0,0 +1,215 @@
...
+			try {
+				this.dataset.writeDataPoint(
+					toAnalyticsEngineDataPoint(
+						observation,
+						payload.resourceAttributes,
+						payload.eventIdentity,
+					),
+				);
+			} catch (error) {
+				console.warn(
```
Rifdhan, line 178: Should this be a full error? Not sure if we expect this to happen in normal operation.

src/metrics/runtime/analytics-engine-sink.ts
```diff
@@ -0,0 +1,215 @@
...
+					`[metrics] Failed to write Analytics Engine data point for ${observation.name}`,
+					error,
+				);
+			}
+		}
+	}
+}
+
+export function createAnalyticsEngineMetricsSink(
+	dataset: unknown,
```
Rifdhan, line 188: Why is this unknown? We should be able to define a type that matches the if-block conditions

src/metrics/runtime/request-metrics.ts
```diff
@@ -25,6 +26,18 @@ export function setMetricsRecorderOnExecutionContext(
...
 
+function createDefaultConfiguredMetricsSinks(
+	env: MetricsEnvLike | undefined,
+	sinks: ConfiguredMetricsSinks,
+): ConfiguredMetricsSinks {
+	return {
+		analyticsEngineSink:
+			sinks.analyticsEngineSink ??
+			createAnalyticsEngineMetricsSink(env?.ANALYTICS),
+		grafanaSink: sinks.grafanaSink,
```
Rifdhan: Why is there a fallback constructor for analytics but not grafana?

src/metrics/runtime/analytics-engine-sink.ts
```diff
@@ -0,0 +1,215 @@
...
+
+export const ANALYTICS_ENGINE_INDEX_FIELDS = [
+	"schema_version",
+	"event_family",
+	"metric_name",
+	"tenant_id",
+	"user_id",
+] as const;
+
+export const ANALYTICS_ENGINE_BLOB_FIELDS = [
```
Rifdhan, line 46: I am a little concerned about maintainability of this mishmash of field names. Some of them seem to be related to tool call properties and some of them are not. How can we ensure this will remain up to date as the code evolves and new tools are added?

src/metrics/runtime/analytics-engine-sink.ts
kanwarbajwa: This is best-effort observability failure at the datapoint level, not a request failure. Using error would overstate the severity and add noisy logs for a path we explicitly isolate from business logic.

src/metrics/runtime/analytics-engine-sink.ts
kanwarbajwa: conversation_id is session/polling-specific and will add it in a later PR with clear semantics

src/metrics/runtime/request-metrics.ts
kanwarbajwa: The real fallback here is the noop sink, not the Analytics constructor itself.

What this branch adds is the default Analytics sink creation path:
`sinks.analyticsEngineSink ?? createAnalyticsEngineMetricsSink(env?.ANALYTICS)`

Grafana looks asymmetric in this PR because its default creation path lives in the separate Grafana PR#126, where we do:
`sinks.grafanaSink ?? createGrafanaOtlpMetricsSink(env)`

So the final design is not meant to special-case Analytics. Both sinks support: explicit injection, default env-driven construction, and then fallback to noop if the sink is still unavailable. The construction inputs differ slightly though: Analytics comes from the native `env.ANALYTICS` binding, while Grafana is derived from endpoint/auth config in `env`.

src/metrics/runtime/analytics-engine-sink.ts
kanwarbajwa: Replaced the inline `unknown` shape check with a dedicated `isAnalyticsEngineDatasetLike(...)` type guard so the binding boundary is typed more clearly

src/metrics/runtime/analytics-engine-sink.ts
kanwarbajwa: Centralized the persisted AE label/resource field mapping and added a guard test so schema drift is caught if approved labels or stored attributes change.

src/metrics/runtime/analytics-engine-sink.ts
Rifdhan: Responded on slack to discuss

Rifdhan (approved):

PR #126, opened 2026-04-29
SCAL-308279 Add Grafana OTLP metrics sink

kanwarbajwa:
## Summary

Add the direct Grafana OTLP/HTTP sink for low-cardinality MCP service metrics. This keeps PR 3 independent from the Analytics Engine sink so it can be reviewed and merged in parallel.

## What Changed

- add a Grafana OTLP metrics sink that emits OTLP JSON over HTTP
- convert counters, gauges, and histograms into OTLP metric datapoints
- normalize OTLP endpoints to `/v1/metrics`
- support Grafana Cloud Basic auth and explicit OTLP auth headers
- wire request metrics to build the Grafana sink from env config
- test payload mapping, endpoint config, auth headers, and failure behavior

## Semantics

The sink sends delta sums and histograms because request-scoped observations are emitted as per-flush deltas. Export failures are surfaced to the recorder, where the existing metrics flush isolation logs and drops them without changing business responses.

### DIFF
--- a/src/metrics/runtime/grafana-otlp-sink.ts
+++ b/src/metrics/runtime/grafana-otlp-sink.ts
@@ -110,52 +110,16 @@ function getProcessEnvValue(name: string): string | undefined {
 
 function readConfigValue(
 	env: GrafanaOtlpEnvLike | undefined,
-	...keys: string[]
+	key: string,
 ): string | undefined {
-	for (const key of keys) {
-		const envValue = env?.[key];
-		if (typeof envValue === "string" && envValue.length > 0) {
-			return envValue;
-		}
-
-		const processEnvValue = getProcessEnvValue(key);
-		if (processEnvValue && processEnvValue.length > 0) {
-			return processEnvValue;
-		}
+	const envValue = env?.[key];
+	if (typeof envValue === "string" && envValue.length > 0) {
+		return envValue;
 	}
 
-	return undefined;
-}
-
-function decodeHeaderValue(value: string): string {
-	try {
-		return decodeURIComponent(value);
-	} catch {
-		return value;
-	}
-}
-
-function readOtlpAuthorizationHeader(
-	env: GrafanaOtlpEnvLike | undefined,
-): string | undefined {
-	const rawHeaders = readConfigValue(env, "OTEL_EXPORTER_OTLP_HEADERS");
-	if (!rawHeaders) {
-		return undefined;
-	}
-
-	for (const header of rawHeaders.split(",")) {
-		const separatorIndex = header.indexOf("=");
-		if (separatorIndex === -1) {
-			continue;
-		}
-
-		const key = header.slice(0, separatorIndex).trim().toLowerCase();
-		if (key !== "authorization") {
-			continue;
-		}
-
-		const value = header.slice(separatorIndex + 1).trim();
-		return value ? decodeHeaderValue(value) : undefined;
+	const processEnvValue = getProcessEnvValue(key);
+	if (processEnvValue && processEnvValue.length > 0) {
+		return processEnvValue;
 	}
 
 	return undefined;
@@ -398,36 +362,16 @@ export function toOtlpMetricsPayload(
 export function resolveGrafanaOtlpSinkConfig(
 	env?: GrafanaOtlpEnvLike,
 ): GrafanaOtlpSinkConfig | undefined {
-	const endpoint = readConfigValue(
-		env,
-		"GRAFANA_OTLP_METRICS_ENDPOINT",
-		"GRAFANA_OTLP_ENDPOINT",
-		"OTEL_EXPORTER_OTLP_METRICS_ENDPOINT",
-		"OTEL_EXPORTER_OTLP_ENDPOINT",
-	);
+	const endpoint = readConfigValue(env, "GRAFANA_OTLP_ENDPOINT");
 	if (!endpoint) {
 		return undefined;
 	}
 
 	return {
 		endpoint: normalizeOtlpMetricsEndpoint(endpoint),
-		username: readConfigValue(
-			env,
-			"GRAFANA_OTLP_USERNAME",
-			"GRAFANA_CLOUD_ACCOUNT_ID",
-			"GRAFANA_CLOUD_INSTANCE_ID",
-		),
-		apiToken: readConfigValue(
-			env,
-			"GRAFANA_OTLP_API_TOKEN",
-			"GRAFANA_CLOUD_API_TOKEN",
-		),
-		authHeader:
-			readConfigValue(
-				env,
-				"GRAFANA_OTLP_AUTH_HEADER",
-				"OTEL_EXPORTER_OTLP_AUTH_HEADER",
-			) ?? readOtlpAuthorizationHeader(env),
+		username: readConfigValue(env, "GRAFANA_OTLP_USERNAME"),
+		apiToken: readConfigValue(env, "GRAFANA_OTLP_API_TOKEN"),
+		authHeader: readConfigValue(env, "GRAFANA_OTLP_AUTH_HEADER"),
 	};
 }
 
--- a/src/metrics/runtime/grafana-otlp-sink.ts
+++ b/src/metrics/runtime/grafana-otlp-sink.ts
@@ -86,6 +86,14 @@ type GrafanaOtlpMetricsSinkOptions = GrafanaOtlpSinkConfig & {
 	fetchFn?: typeof fetch;
 };
 
+type AggregatedMetricDataPoint = {
+	attributes: OtlpAttribute[];
+	bucketCounts: number[];
+	count: number;
+	timestampMs: number;
+	value: number;
+};
+
 const OTLP_SCOPE_NAME = "thoughtspot.mcp.metrics.runtime";
 const OTLP_AGGREGATION_TEMPORALITY_DELTA = 1;
 const JSON_HEADERS = {
@@ -187,16 +195,16 @@ function toAttributes(
 }
 
 function toTimeUnixNano(timestampMs: number): string {
-	return String(Math.trunc(timestampMs * 1_000_000));
+	return (BigInt(Math.trunc(timestampMs * 1_000)) * 1_000n).toString();
 }
 
-function getHistogramBucketCounts(value: number): string[] {
+function getHistogramBucketCounts(value: number): number[] {
 	const bucketCounts = new Array(HISTOGRAM_BUCKETS_MS.length + 1).fill(
 		0,
 	) as number[];
 	const bucketIndex = HISTOGRAM_BUCKETS_MS.findIndex((bound) => value <= bound);
 	bucketCounts[bucketIndex === -1 ? bucketCounts.length - 1 : bucketIndex] = 1;
-	return bucketCounts.map(String);
+	return bucketCounts;
 }
 
 function groupObservationsByMetric(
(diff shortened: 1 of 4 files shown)


src/metrics/runtime/grafana-otlp-sink.ts
```diff
@@ -0,0 +1,395 @@
...
+		case "histogram":
+			return {
+				name,
+				histogram: {
+					aggregationTemporality: OTLP_AGGREGATION_TEMPORALITY_DELTA,
+					dataPoints: observations.map(toHistogramDataPoint),
+				},
+			};
+	}
+}
```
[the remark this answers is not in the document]
kanwarbajwa: Addressed by aggregating observations before OTLP conversion so each Metric has one datapoint per unique attribute set. Counters are summed, histograms aggregate count/sum/buckets, and gauges keep the latest value. The recorder already sets `flushed = true` before snapshot/await, and the sink awaits `fetch` and error body reads.

src/metrics/runtime/grafana-otlp-sink.ts
```diff
@@ -0,0 +1,395 @@
...
+			if (value === undefined || value === "") {
+				return [];
+			}
+			return [{ key, value: toOtlpValue(value) }];
+		});
+}
+
+function toTimeUnixNano(timestampMs: number): string {
+	return String(Math.trunc(timestampMs * 1_000_000));
+}
```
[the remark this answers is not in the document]
kanwarbajwa: Timestamp conversion now uses `BigInt` for the nanosecond string instead of multiplying into an unsafe JS number range. Added regression coverage for fractional millisecond timestamps.

kanwarbajwa:
@gemini-code-assist please review again

src/metrics/runtime/grafana-otlp-sink.ts
```diff
@@ -0,0 +1,451 @@
...
+	const aggregated = new Map<string, AggregatedMetricDataPoint>();
+
+	for (const observation of observations) {
+		const attributes = toAttributes(observation.labels);
+		const attributeSetKey = getAttributeSetKey(attributes);
+		const dataPoint = aggregated.get(attributeSetKey) ?? {
+			attributes,
+			bucketCounts: new Array(HISTOGRAM_BUCKETS_MS.length + 1).fill(
+				0,
+			) as number[],
```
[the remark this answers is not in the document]
kanwarbajwa: `bucketCounts` is now only allocated for histogram datapoints; counters and gauges use an empty array and avoid the unnecessary allocation.

kanwarbajwa:
@gemini-code-assist please review again

src/metrics/runtime/grafana-otlp-sink.ts
```diff
@@ -0,0 +1,452 @@
...
+
+function encodeBase64(value: string): string {
+	if (typeof btoa === "function") {
+		return btoa(value);
+	}
+	if (typeof Buffer !== "undefined") {
+		return Buffer.from(value, "utf8").toString("base64");
+	}
+	throw new Error("No base64 encoder is available for Grafana OTLP auth");
+}
```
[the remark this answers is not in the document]
kanwarbajwa: `Buffer` remains preferred where available, and the `btoa` fallback now explicitly UTF-8 encodes via `TextEncoder` before base64 conversion.

src/metrics/runtime/grafana-otlp-sink.ts
```diff
@@ -0,0 +1,452 @@
...
+			const value = attributes[key];
+			if (value === undefined || value === "") {
+				return [];
+			}
+			return [{ key, value: toOtlpValue(value) }];
+		});
+}
+
+function toTimeUnixNano(timestampMs: number): string {
+	return (BigInt(Math.trunc(timestampMs * 1_000)) * 1_000n).toString();
```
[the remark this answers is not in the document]
kanwarbajwa: Timestamp conversion now splits integer milliseconds from fractional milliseconds before `BigInt` conversion, avoiding unsafe epoch-scale floating-point multiplication. The fractional part is kept at microsecond precision.

src/metrics/runtime/grafana-otlp-sink.ts
```diff
@@ -0,0 +1,452 @@
...
+	for (const observation of observations) {
+		const metricObservations = grouped.get(observation.name) ?? [];
+		metricObservations.push(observation);
+		grouped.set(observation.name, metricObservations);
+	}
+	return grouped;
+}
+
+function getAttributeSetKey(attributes: readonly OtlpAttribute[]): string {
+	return JSON.stringify(attributes);
```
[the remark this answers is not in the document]
kanwarbajwa: I’m leaving this as-is for now. `JSON.stringify` is stable here because attributes are sorted, and the sink flushes request-scoped batches rather than high-volume background aggregation. If we later reuse this for larger buffered batches, replacing the key with a cheaper custom serializer/hash would be reasonable.

src/metrics/runtime/grafana-otlp-sink.ts
```diff
@@ -0,0 +1,452 @@
...
+
+		const response = await this.fetchFn(this.endpoint, {
+			method: "POST",
+			headers: this.headers,
+			body: JSON.stringify(toOtlpMetricsPayload(payload)),
+		});
+
+		if (!response.ok) {
+			throw new Error(
+				`Grafana OTLP metrics export failed with status ${response.status}: ${await response.text()}`,
```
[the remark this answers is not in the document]
kanwarbajwa: Capped Grafana OTLP export error body text.

kanwarbajwa:
@gemini-code-assist please review again

src/metrics/runtime/request-metrics.ts
```diff
@@ -34,12 +38,25 @@ export function clearMetricsRecorderFromExecutionContext(
 	delete (ctx as MetricsExecutionContext)[METRICS_RECORDER_SYMBOL];
 }
 
+function createDefaultConfiguredMetricsSinks(
+	env: MetricsEnvLike | undefined,
+	sinks: ConfiguredMetricsSinks,
+): ConfiguredMetricsSinks {
+	return {
+		analyticsEngineSink: sinks.analyticsEngineSink,
+		grafanaSink: sinks.grafanaSink ?? createGrafanaOtlpMetricsSink(env),
```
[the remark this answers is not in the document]
kanwarbajwa: using a module-level `WeakMap` keyed by the Worker `env` object. This avoids rebuilding the default Grafana sink on each request within a warm isolate without using secret values as cache keys or forcing a global singleton. The cache is opportunistic per isolate; cold starts or new isolates rebuild the sink.

kanwarbajwa:
@gemini-code-assist please review again

src/metrics/runtime/grafana-otlp-sink.ts
```diff
@@ -0,0 +1,472 @@
...
+}
+
+function getHistogramBucketCounts(value: number): number[] {
+	const bucketCounts = new Array(HISTOGRAM_BUCKETS_MS.length + 1).fill(
+		0,
+	) as number[];
+	const bucketIndex = HISTOGRAM_BUCKETS_MS.findIndex((bound) => value <= bound);
+	bucketCounts[bucketIndex === -1 ? bucketCounts.length - 1 : bucketIndex] = 1;
+	return bucketCounts;
+}
```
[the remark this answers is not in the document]
kanwarbajwa: Done

src/metrics/runtime/grafana-otlp-sink.ts
```diff
@@ -0,0 +1,472 @@
...
+			case "gauge":
+				dataPoint.value = observation.value;
+				dataPoint.count = 1;
+				dataPoint.timestampMs = observation.timestampMs;
+				break;
+			case "histogram": {
+				const bucketCounts = getHistogramBucketCounts(observation.value);
+				for (const [index, bucketCount] of bucketCounts.entries()) {
+					dataPoint.bucketCounts[index] += bucketCount;
+				}
```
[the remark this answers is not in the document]
kanwarbajwa: Done

kanwarbajwa:
@gemini-code-assist please review again

src/metrics/runtime/grafana-otlp-sink.ts
```diff
@@ -0,0 +1,466 @@
...
+	attributes: OtlpAttribute[];
+	bucketCounts: number[];
+	count: number;
+	timestampMs: number;
+	value: number;
+};
+
+const OTLP_SCOPE_NAME = "thoughtspot.mcp.metrics.runtime";
+const OTLP_AGGREGATION_TEMPORALITY_DELTA = 1;
+const MAX_EXPORT_ERROR_BODY_LENGTH = 1_000;
```
Rifdhan: Nit: why the underscore notation?

src/metrics/runtime/grafana-otlp-sink.ts
```diff
@@ -0,0 +1,466 @@
...
+		return undefined;
+	}
+	return process.env?.[name];
+}
+
+function readConfigValue(
+	env: GrafanaOtlpEnvLike | undefined,
+	...keys: string[]
+): string | undefined {
+	for (const key of keys) {
```
Rifdhan: This feels odd, why do we have multiple possible keys for the same value?

src/metrics/runtime/grafana-otlp-sink.ts
```diff
@@ -0,0 +1,466 @@
...
+		let binary = "";
+		for (const byte of bytes) {
+			binary += String.fromCharCode(byte);
+		}
+		return btoa(binary);
+	}
+	throw new Error("No base64 encoder is available for Grafana OTLP auth");
+}
+
+function toOtlpValue(value: MetricLabelValue | string): OtlpAttributeValue {
```
Rifdhan: MetricLabelValue already includes string as a possible type, why do we need `| string` again here? Same for other places where this type is used.

src/metrics/runtime/grafana-otlp-sink.ts
```diff
@@ -0,0 +1,466 @@
...
+		return { doubleValue: value };
+	}
+	return { stringValue: value };
+}
+
+function toAttributes(
+	attributes: Record<string, MetricLabelValue | string | undefined>,
+): OtlpAttribute[] {
+	return Object.keys(attributes)
+		.sort()
```
Rifdhan, line 157: Why the sort?

src/metrics/runtime/grafana-otlp-sink.ts
```diff
@@ -0,0 +1,466 @@
...
+	if (typeof value === "boolean") {
+		return { boolValue: value };
+	}
+	if (typeof value === "number") {
+		return { doubleValue: value };
+	}
+	return { stringValue: value };
+}
+
+function toAttributes(
```
Rifdhan: Nit: would call it toOtlpAttributes for clarity, since we already pass attributes in, and it matches other names like toOtlpValue

src/metrics/runtime/grafana-otlp-sink.ts
```diff
@@ -0,0 +1,466 @@
...
+		.flatMap((key) => {
+			const value = attributes[key];
+			if (value === undefined || value === "") {
+				return [];
+			}
+			return [{ key, value: toOtlpValue(value) }];
+		});
+}
+
+function toTimeUnixNano(timestampMs: number): string {
```
Rifdhan, line 167: See Slack comment

src/metrics/runtime/grafana-otlp-sink.ts
```diff
@@ -0,0 +1,466 @@
...
+			};
+	}
+}
+
+function normalizeOtlpMetricsEndpoint(endpoint: string): string {
+	const trimmed = endpoint.trim().replace(/\/+$/, "");
+	return trimmed.endsWith("/v1/metrics") ? trimmed : `${trimmed}/v1/metrics`;
+}
+
+function buildAuthorizationHeader(
```
Rifdhan: buildAuthorizationHeaderValue, since it's just the value, to differentiate from the naming of buildRequestHeaders which provides key-values

src/metrics/runtime/grafana-otlp-sink.ts
```diff
@@ -0,0 +1,466 @@
...
+		? { ...JSON_HEADERS, Authorization: authorization }
+		: JSON_HEADERS;
+}
+
+async function getExportErrorBody(response: Response): Promise<string> {
+	const text = await response.text();
+	if (text.length <= MAX_EXPORT_ERROR_BODY_LENGTH) {
+		return text;
+	}
+	return `${text.slice(0, MAX_EXPORT_ERROR_BODY_LENGTH)}...`;
```
Rifdhan: Technically this is 3 chars longer than the max length, should we truncate to accommodate the ellipses?

src/metrics/runtime/request-metrics.ts
```diff
@@ -37,12 +40,44 @@ export function clearMetricsRecorderFromExecutionContext(
 	delete (ctx as MetricsExecutionContext)[METRICS_RECORDER_SYMBOL];
 }
 
+function getCachedGrafanaSink(
```
Rifdhan, line 44: Can we have multiple different env values at once? Wondering why we need an env-based mapping

src/metrics/runtime/grafana-otlp-sink.ts
kanwarbajwa: The sort is deliberate. We aggregate by serialized attribute set, so we need deterministic ordering; otherwise equivalent labels in different insertion orders would produce different aggregation keys.

src/metrics/runtime/grafana-otlp-sink.ts
kanwarbajwa: Updated the truncation so the configured max includes the ellipsis.

src/metrics/runtime/grafana-otlp-sink.ts
kanwarbajwa: Renamed to `buildAuthorizationHeaderValue`

src/metrics/runtime/grafana-otlp-sink.ts
kanwarbajwa: Addressed. We now round the fractional millisecond portion directly to nanoseconds and handle carry into the integer millisecond part.

src/metrics/runtime/grafana-otlp-sink.ts
kanwarbajwa: Renamed to `toOtlpAttributes` for clarity.

src/metrics/runtime/grafana-otlp-sink.ts
kanwarbajwa: Removed the redundant `| string` usages.

src/metrics/runtime/grafana-otlp-sink.ts
kanwarbajwa: The original intent was to be flexible and accept both our own Grafana env names and OTLP-style names, but that makes the config surface harder to reason about. Fixed this to have single key value

src/metrics/runtime/request-metrics.ts
kanwarbajwa: Probably not in normal production traffic, but I’d keep it this way. The main goal is to avoid rebuilding the sink on every request while still scoping reuse to the same env/config object. In production that will usually behave like a single cached sink anyway, but keying by `env` avoids accidental cross-env reuse in tests, local dev.

src/metrics/runtime/grafana-otlp-sink.ts
Rifdhan: Would be good to leave a comment noting this, so future devs don't remove the sort thinking it's not needed

src/metrics/runtime/request-metrics.ts
Rifdhan: Might be good to note this in a comment

Rifdhan (approved):

PR #129, opened 2026-04-30
SCAL-309983 Harden metrics request isolation

kanwarbajwa:
## Summary

Harden the metrics request wrapper so metrics setup, scheduling, and flush failures cannot
block MCP business logic or become visible to MCP clients.

## What Changed

- fall back to a noop sink if metrics config or sink construction throws
- keep `recorder.flush()` focused on metrics delivery, not Worker scheduling
- schedule request metrics flush through `ctx.waitUntil(...)` without awaiting delivery
- log waitUntil scheduling failures separately from metrics execution failures
- add tests for setup fallback, slow flushes, handler failures, waitUntil failures, and
  delivery failures

## Semantics

Direct `recorder.flush()` calls still await delivery. `withRequestMetrics(...)` only schedules
the request-scoped flush so slow, failing, or unavailable sinks do not delay responses.

Metrics failures remain best-effort side effects and are never propagated to MCP clients.

### DIFF
--- a/src/metrics/runtime/request-metrics.ts
+++ b/src/metrics/runtime/request-metrics.ts
@@ -2,6 +2,7 @@ import {
 	type MetricsRecorder,
 	RequestMetricsRecorder,
 } from "./metrics-recorder";
+import { NoopMetricsSink } from "./noop-sink";
 import {
 	type ConfiguredMetricsSinks,
 	type MetricsEnvLike,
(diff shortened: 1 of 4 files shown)


src/metrics/runtime/request-metrics.ts
```diff
@@ -41,13 +42,40 @@ export function createRequestMetricsRecorder(
...
+	recorder: MetricsRecorder,
+	ctx: ExecutionContext,
+): void {
+	try {
+		recorder.flush(ctx).catch((error) => {
+			console.error(
+				"[metrics] Failed to schedule request metrics flush",
+				error,
+			);
+		});
```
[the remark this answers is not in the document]
kanwarbajwa: Updated this to separate metrics delivery from Worker scheduling. `recorder.flush()` now only handles delivery, while `withRequestMetrics(...)` creates the guarded flush promise and schedules it with `ctx.waitUntil(...)`.

That lets us distinguish scheduling failures from delivery failures, and both remain metrics-only side effects. We never await metrics delivery on the MCP request path, and metrics failures are not propagated to MCP clients.

src/metrics/runtime/request-metrics.ts
```diff
@@ -41,13 +42,45 @@ export function createRequestMetricsRecorder(
...
+		return new RequestMetricsRecorder({
+			sink,
+			resourceAttributes: config.resourceAttributes,
+		});
+	} catch (error) {
+		console.error(
+			"[metrics] Failed to initialize request metrics recorder; using noop sink",
+			error,
+		);
+		return new RequestMetricsRecorder({
```
Rifdhan: There's no chance this can also fail?

Rifdhan (approved):

src/metrics/runtime/request-metrics.ts
kanwarbajwa: Fair point. In practice the previous fallback was effectively safe because both `RequestMetricsRecorder` and `NoopMetricsSink` have trivial constructors, but it still meant the fail-open path was constructing another recorder and in theory it could fail if in future we changed the constructors etc...

I changed this so `createRequestMetricsRecorder()` now returns the `MetricsRecorder` interface and, on initialization failure, returns a shared stateless `NOOP_METRICS_RECORDER` directly. That removes the second construction step from the fallback path and makes the request-isolation cleaner.

Rifdhan (approved):

PR #132, opened 2026-05-01
Release 2026-05-01 version

Rifdhan:
- Add new entry in version registry for 2026-05-01
- Add new "latest" version option, mapping to newest available non-beta version
- Remove "default" version, default will now be latest
- Add "backwards-compatibility-default" version which points to older version for now, will be removed in the future
- Add metrics tracking of versioning
- Update tests

### DIFF
--- a/src/servers/mcp-server.ts
+++ b/src/servers/mcp-server.ts
@@ -9,7 +9,7 @@ import { TrackEvent } from "../metrics";
 import { WithSpan } from "../metrics/tracing/tracing-utils";
 import { context, SpanStatusCode, trace } from "@opentelemetry/api";
 import { BaseMCPServer, type Context } from "./mcp-server-base";
-import { resolveApiVersion } from "./version-registry";
+import { resolveApiVersion, type VersionConfig } from "./version-registry";
 import {
 	GetRelevantQuestionsSchema,
 	GetAnswerSchema,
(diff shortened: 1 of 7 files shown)


src/servers/mcp-server.ts
```diff
@@ -32,9 +32,27 @@ export class MCPServer extends BaseMCPServer {
...
+
 		// Resolve the API version to get the appropriate tool configuration
-		const versionConfig = resolveApiVersion(this.ctx.props.apiVersion);
+		let versionConfig: VersionConfig;
+		try {
+			versionConfig = resolveApiVersion(this.ctx.props.apiVersion);
+		} catch (error) {
+			console.error("Error resolving API version, using default:", error);
+			versionConfig = resolveApiVersion();
+		}
```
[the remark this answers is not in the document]
Rifdhan: Added

src/servers/mcp-server.ts
```diff
@@ -32,9 +32,28 @@ export class MCPServer extends BaseMCPServer {
...
+		span?.setAttribute(
+			"api_version_resolved",
+			// The plain date will be the last entry if multiple labels
+			versionConfig.version[versionConfig.version.length - 1],
+		);
 
 		// Get base tools from version config
 		let tools = [...versionConfig.tools];
```
kanwarbajwa, line 59: `listTools()` is version-aware, but `callTool()` is not. This still allows clients that already know the V2 tool names to invoke them through `/mcp` or `/bearer/mcp`, even though those routes are supposed to stay on the old tool surface. We should gate tool execution by the resolved version too, not just tool discovery.

src/servers/mcp-server.ts
```diff
@@ -32,9 +32,28 @@ export class MCPServer extends BaseMCPServer {
...
+
 		// Resolve the API version to get the appropriate tool configuration
-		const versionConfig = resolveApiVersion(this.ctx.props.apiVersion);
+		let versionConfig: VersionConfig;
+		try {
+			versionConfig = resolveApiVersion(this.ctx.props.apiVersion);
+		} catch (error) {
+			console.error("Error resolving API version, using default:", error);
+			span?.recordException(error as Error);
+			versionConfig = resolveApiVersion();
```
kanwarbajwa, line 50: Invalid or unknown `api-version` values currently end up on `latest`. `resolveApiVersion()` throws, but this catch retries with no version, which now defaults to `latest`. Per the versioning doc, `/mcp` should stay on the old tools unless the caller explicitly requests `beta` or a valid release date. A typo like `?api-version=garbage` should not silently upgrade the caller to the latest surface.

src/servers/mcp-server.ts
Rifdhan: Technically that's possible but a host agent would not know about a tool that is not returned in listTools() so it should never get called. The versioning system is not a security or access control restriction, so I feel it does not need to be bulletproof in that regard.

src/servers/mcp-server.ts
Rifdhan: I wanted the design to reflect the eventual design which we will have in 3 months, after this one-time transition process. That design will be latest-by-default. For backwards compatibility, we have stubs in place which use backwards-compatible-default, and that is meant to provide backwards compatibility for older clients who are not passing an api_version flag at all to continue seeing the same tools they have been seeing. I don't think it's necessary to handle "garbage" version flags because that is not the use case for backwards compatibility. I did explore whether we can return an error on invalid flag values, but the listTools MCP spec does not really support this, and in reality it causes the call to silently fail the host agent to not know or understand what went wrong.

kanwarbajwa (approved):

PR #143, opened 2026-05-07
SCAL-308282 Track polling and first response latency

kanwarbajwa:
## Summary

Instrument the analysis polling bridge so we can measure backend readiness, client polling behavior, and sessions that never get polled.

## What Changed

- add shared analysis metric helpers for session counters, poll wait, first buffered update, first poll delay, first non-empty response, never-polled sessions, and storage errors
- persist timing metadata in the ConversationStorageServer Durable Object and expose it through the storage client so one-shot lifecycle metrics are emitted once per response
- record request-scoped analysis counters and poll-wait histograms in the MCP server analysis tool methods
- add focused specs for the conversation storage timing state and the analysis tool metrics flow

## Notes

- first-buffered, first-poll, first-non-empty, and never-polled metrics are emitted from the storage Durable Object because it owns the cross-request conversation state
- local design docs were updated for our internal reference but are not part of this commit

### DIFF
--- a/src/metrics/runtime/analysis-metrics.ts
+++ b/src/metrics/runtime/analysis-metrics.ts
@@ -1,0 +1,163 @@
+import {
+	METRIC_NAMES,
+	type MetricLabelInput,
+	type MetricOutcome,
+} from "./metric-types";
+import type { MetricsRecorder } from "./metrics-recorder";
+
+export const STREAM_STORAGE_OPERATIONS = {
+	initialize: "initialize",
+	append: "append",
+	messages: "messages",
+	alarm: "alarm",
+	unknown: "unknown",
+} as const;
+
+export type StreamStorageOperation =
+	(typeof STREAM_STORAGE_OPERATIONS)[keyof typeof STREAM_STORAGE_OPERATIONS];
+
+function recordCountAnalysisMetric(
+	recorder: MetricsRecorder | undefined,
+	metricName: string,
+	outcome: MetricOutcome,
+	extraLabels?: MetricLabelInput,
+): void {
+	if (!recorder) {
+		return;
+	}
+
+	recorder.count(metricName, 1, {
+		outcome,
+		...extraLabels,
+	});
+}
+
+function recordHistogramAnalysisMetric(
+	recorder: MetricsRecorder | undefined,
+	metricName: string,
+	durationMs: number,
+	outcome: MetricOutcome,
+	extraLabels?: MetricLabelInput,
+): void {
+	if (!recorder) {
+		return;
+	}
+
+	recorder.histogram(metricName, durationMs, {
+		outcome,
+		...extraLabels,
+	});
+}
+
+export function recordAnalysisSessionCreatedMetric(
+	recorder: MetricsRecorder | undefined,
+	outcome: MetricOutcome,
+): void {
+	recordCountAnalysisMetric(
+		recorder,
+		METRIC_NAMES.analysisSessionsCreatedTotal,
+		outcome,
+	);
+}
+
+export function recordAnalysisMessageSentMetric(
+	recorder: MetricsRecorder | undefined,
+	outcome: MetricOutcome,
+): void {
+	recordCountAnalysisMetric(
+		recorder,
+		METRIC_NAMES.analysisMessagesSentTotal,
+		outcome,
+	);
+}
+
+export function recordAnalysisUpdatesPolledMetric(
+	recorder: MetricsRecorder | undefined,
+	outcome: MetricOutcome,
+): void {
+	recordCountAnalysisMetric(
+		recorder,
+		METRIC_NAMES.analysisUpdatesPolledTotal,
+		outcome,
+	);
+}
+
+export function recordAnalysisPollWaitMetric(
+	recorder: MetricsRecorder | undefined,
+	durationMs: number,
+	outcome: MetricOutcome,
+	isDone: boolean,
+): void {
+	recordHistogramAnalysisMetric(
+		recorder,
+		METRIC_NAMES.analysisPollWaitMs,
+		durationMs,
+		outcome,
+		{
+			is_done: isDone,
+		},
+	);
+}
+
+export function recordAnalysisFirstBufferedUpdateMetric(
+	recorder: MetricsRecorder | undefined,
+	durationMs: number,
+	outcome: MetricOutcome,
+): void {
+	recordHistogramAnalysisMetric(
+		recorder,
+		METRIC_NAMES.analysisFirstBufferedUpdateMs,
+		durationMs,
+		outcome,
+	);
+}
+
+export function recordAnalysisFirstPollDelayMetric(
+	recorder: MetricsRecorder | undefined,
+	durationMs: number,
+	outcome: MetricOutcome,
+): void {
+	recordHistogramAnalysisMetric(
+		recorder,
+		METRIC_NAMES.analysisFirstPollDelayMs,
+		durationMs,
+		outcome,
+	);
+}
+
+export function recordAnalysisFirstNonEmptyResponseMetric(
+	recorder: MetricsRecorder | undefined,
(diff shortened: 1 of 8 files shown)


src/servers/conversation-storage-server.ts
```diff
@@ -159,6 +259,62 @@ export class ConversationStorageServer {
...
+		}
+
+		const messages: Message[] = [];
+		const messagesMap = await this.getInBatches<Message>(keys);
+		for (let i = readBookmark; i < writeBookmark; i++) {
+			const message = messagesMap.get(MESSAGE_KEY_PREFIX + i);
+			if (message) {
+				messages.push(message);
+			}
+		}
```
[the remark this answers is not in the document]
kanwarbajwa: extracted the shared message-range read path into `getMessagesInRange(...)` so `getNewMessagesAndUpdateBookmark()` and `recordFirstNonEmptyResponse()` no longer duplicate the same bookmark/key walk.

src/metrics/runtime/analysis-metrics.ts
```diff
@@ -0,0 +1,164 @@
...
+	messages: "messages",
+	firstNonEmptyResponse: "first_non_empty_response",
+	alarm: "alarm",
+	unknown: "unknown",
+} as const;
+
+export type StreamStorageOperation =
+	(typeof STREAM_STORAGE_OPERATIONS)[keyof typeof STREAM_STORAGE_OPERATIONS];
+
+function countAnalysisMetric(
```
Rifdhan: This and the next fn don't have "record" prefix, while the rest of the fns in the file do have it. Is this intentional, or can we add it for these too for consistency?

src/servers/mcp-server.ts
```diff
@@ -463,46 +529,99 @@ Provide this url to the user as a link to view the liveboard in ThoughtSpot.`;
...
+					newMessagesState.responseStartedAtMs ??
+					messagesState.responseStartedAtMs;
+				messagesState.firstBufferedUpdateAtMs =
+					newMessagesState.firstBufferedUpdateAtMs ??
+					messagesState.firstBufferedUpdateAtMs;
+				messagesState.firstPollAtMs =
+					newMessagesState.firstPollAtMs ?? messagesState.firstPollAtMs;
+				messagesState.firstNonEmptyResponseAtMs =
+					newMessagesState.firstNonEmptyResponseAtMs ??
+					messagesState.firstNonEmptyResponseAtMs;
```
Rifdhan: I don't feel great about blending business logic data and internal metric data into a single object with no clear distinction. Would prefer to keep the metrics data in a separate object for clarity and isolation.

src/servers/conversation-storage-server.ts
```diff
@@ -56,10 +91,21 @@ export class ConversationStorageServerSQLite {
 					return Response.json(state);
 				}
 
+				case "POST /first-non-empty-response": {
```
Rifdhan: Having an entire new endpoint that is just for a metrics field seems overkill. Can we achieve this completely inside conversation-storage-server by detecting when get messages is returning a non-empty message for the first time?

src/metrics/runtime/analysis-metrics.ts
kanwarbajwa: Added for consistency

src/servers/mcp-server.ts
kanwarbajwa: simplified this so `getNewMessages(...)` returns plain `StreamingMessagesState` again, and the timing milestones stay internal to the conversation storage DO instead of being mixed into the client-facing snapshot.

src/servers/conversation-storage-server.ts
kanwarbajwa: moved first-non-empty detection into `getNewMessagesAndUpdateBookmark()` and removed the extra `POST /first-non-empty-response` endpoint, the storage client method, and the follow-up call from `get_session_updates`.

src/servers/conversation-storage-server.ts
```diff
@@ -60,6 +92,7 @@ export class ConversationStorageServerSQLite {
 					return new Response("Not Found", { status: 404 });
 			}
 		} catch (err) {
+			await this.recordStorageErrorMetric(this.toStorageOperation(operation));
```
Rifdhan, line 95: Nit: prefer to do this after the console.error line, for two reasons:
1. Logging error in console is always reliable and we should do it first before other more complex error handling which can introduce further failure modes resulting in non-logging
2. This is async, so I want to log the error immediately first and foremost

src/servers/conversation-storage-server.ts
```diff
@@ -221,26 +312,142 @@ export class ConversationStorageServerSQLite {
...
+			});
+		} catch (error) {
+			console.error(
+				`[metrics] Failed to record conversation storage error metric for ${this.conversationId}`,
+				error,
+			);
+		}
+	}
+
+	private toStorageOperation(operation: string): StreamStorageOperation {
```
Rifdhan, line 441: Can be static

src/servers/conversation-storage-server.ts
```diff
@@ -221,26 +312,142 @@ export class ConversationStorageServerSQLite {
...
+		switch (operation) {
+			case "initialize":
+				return STREAM_STORAGE_OPERATIONS.initialize;
+			case "append":
+				return STREAM_STORAGE_OPERATIONS.append;
+			case "messages":
+				return STREAM_STORAGE_OPERATIONS.messages;
+			default:
+				return STREAM_STORAGE_OPERATIONS.unknown;
```
Rifdhan, line 450: It looks like we are never recording STREAM_STORAGE_OPERATIONS.alarm, can we remove it?

src/storage-service/storage-service.ts
```diff
@@ -1,16 +1,22 @@
...
- *   GET   /storage/<storageId>/messages   —> getNewMessagesAndUpdateBookmark
+ *   POST  /storage/<conversationId>/initialize —> initializeConversation
+ *   POST  /storage/<conversationId>/append     —> appendMessagesAndRestartTtl
+ *   GET   /storage/<conversationId>/messages   —> getNewMessagesAndUpdateBookmark
  *
- * The storageId is derived by taking a hash of the user's access token and combining it with the
- * conversationId, to ensure no users can access each other's conversations.
+ * The route path uses the conversation id. Isolation between users happens at the
+ * Durable Object id level by hashing the access token and combining it with the
+ * conversation id when constructing the stub name.
```
Rifdhan, line 19: Why did we change these header comments? We are still using the storageId concept from my last change, there should be no change here.

src/servers/conversation-storage-server.ts
```diff
@@ -41,7 +69,11 @@ export class ConversationStorageServerSQLite {
 		try {
 			switch (`${request.method} /${operation}`) {
 				case "POST /initialize": {
-					await this.initializeConversation();
+					const body =
+						((await request
+							.json()
+							.catch(() => ({}))) as Partial<ConversationMetricsState>) ?? {};
```
Rifdhan, line 75: Let's update the top-level header comment of this class to mention that we are passing metrics data in the request body

src/servers/conversation-storage-server.ts
```diff
@@ -96,7 +143,10 @@ export class ConversationStorageServerSQLite {
 		newMessages: Message[],
 		isDone = false,
 	): Promise<void> {
-		const existingIsDone = await this.state.storage.get<boolean>(IS_DONE_KEY);
+		const [existingIsDone, metricsState] = await Promise.all([
+			this.state.storage.get<boolean>(IS_DONE_KEY),
+			this.getConversationMetricsState(),
+		]);
```
Rifdhan, line 149: Let's just call getConversationStateSnapshot() here and do away with getConversationMetricsState(). It feels unclean to be fetching 2 different keys in separate transactions.

PR #155, opened 2026-05-19
Remove logging of potentially sensitive information

Rifdhan:
- Update tests
- Remove some additional dead code

### DIFF
--- a/src/streaming-utils.ts
+++ b/src/streaming-utils.ts
@@ -69,7 +69,6 @@ export const processSendAgentConversationMessageStreamingResponse = async (
 						if (!line.startsWith("data: ")) {
 							console.warn(
 								"Unknown line in event stream, does not start with 'data:'",
-								`"${line}"`,
 							);
 							continue;
 						}
@@ -135,7 +134,10 @@ export const processSendAgentConversationMessageStreamingResponse = async (
 								// We intentionally ignore the above events
 								nMessagesIgnored++;
 							} else if (item.type === "error") {
-								console.error("Error event in event stream: ", item);
+								console.error(
+									"Error event in event stream, error code",
+									item.error_code,
+								);
 								recordUpstreamStreamMessageMetric(
 									recorder,
 									upstreamOperation,
@@ -145,7 +147,7 @@ export const processSendAgentConversationMessageStreamingResponse = async (
 								spanHasError = true;
 								span.setStatus({
 									code: SpanStatusCode.ERROR,
-									message: item,
+									message: `Error event in event stream, error code: ${item.error_code}`,
 								});
 								newMessages.push({
 									is_thinking: false,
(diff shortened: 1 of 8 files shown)


src/streaming-utils.ts
```diff
@@ -145,15 +144,15 @@ export const processSendAgentConversationMessageStreamingResponse = async (
...
+									message: "Error event in event stream",
 								});
 								newMessages.push({
 									is_thinking: false,
 									type: "text",
 									text: item.display_message || "Something went wrong",
 								});
 							} else {
-								console.warn("Unknown event in event stream: ", item);
+								console.warn("Unknown event in event stream:", item.type);
```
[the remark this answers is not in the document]
Rifdhan: We already check item.type in the if-blocks before this else block, so this is not a concern

src/streaming-utils.ts
```diff
@@ -135,7 +134,7 @@ export const processSendAgentConversationMessageStreamingResponse = async (
 								// We intentionally ignore the above events
 								nMessagesIgnored++;
 							} else if (item.type === "error") {
-								console.error("Error event in event stream: ", item);
+								console.error("Error event in event stream");
```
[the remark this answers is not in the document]
Rifdhan: Added

src/streaming-utils.ts
```diff
@@ -145,15 +144,15 @@ export const processSendAgentConversationMessageStreamingResponse = async (
 								spanHasError = true;
 								span.setStatus({
 									code: SpanStatusCode.ERROR,
-									message: item,
+									message: "Error event in event stream",
```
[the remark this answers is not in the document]
Rifdhan: Added

mouryabalabhadra:
@copilot Please review PR

src/thoughtspot/thoughtspot-service.ts
```diff
@@ -243,14 +243,6 @@ export class ThoughtSpotService {
 				generation_number,
 			});
 
-			console.log(
-				"[DEBUG] Getting Data for session_identifier: ",
-				session_identifier,
-				"generation_number: ",
-				generation_number,
-				"instanceUrl: ",
-				(this.client as any).instanceUrl,
```
kanwarbajwa, line 252: The only PII I see here is instanceUrl, which we can replace with tenantId

static/oauth-callback.js
```diff
@@ -155,7 +152,6 @@
             throw new Error(`Failed to store token (Status: ${storeResponse.status}): ${errorText}`);
         }
 
-        console.log('Redirecting to:', responseData.redirectTo);
```
kanwarbajwa, line 158: Should we add tenantId here? Redirecting to tenantId

src/thoughtspot/thoughtspot-service.ts
Rifdhan: I feel this log is not useful anyway, we don't need to debug anything with this flow and it is also in the old V1 tool so I feel we can just remove it

static/oauth-callback.js
Rifdhan: This is logged in the end user's browser while going through OAuth, I don't think we have any need to log it at all

kanwarbajwa (approved):

PR #165, opened 2026-06-18
Initial commit to move to common package

adityamittal3107:


src/index.ts
```diff
@@ -36,78 +44,132 @@ const config: ResolveConfigFn = (env: Env, _trigger) => {
...
+	},
+};
+
+const oauthFetchHandler = createOAuthHandler<Props>({
+	serverInfo: {
+		name: "ThoughtSpot Spotter",
+		logo: "https://avatars.githubusercontent.com/u/8906680?s=200&v=4",
+		description: "MCP Server for ThoughtSpot Agent",
+	},
+	mcpServerClass: ThoughtSpotMCP as unknown as Parameters<
```
Rifdhan: Is there nothing better we can do for types?

src/index.ts
```diff
@@ -36,78 +44,132 @@ const config: ResolveConfigFn = (env: Env, _trigger) => {
...
+		typeof createOAuthHandler
+	>[0]["mcpServerClass"],
+	// Cast needed: pkg AuthHooks<BaseProps> has a narrower clientName than local Props.
+	hooks: hooks as any,
```
Rifdhan: Can hooks be a well defined type contract from the auth package?

src/index.ts
```diff
@@ -36,78 +44,132 @@ const config: ResolveConfigFn = (env: Env, _trigger) => {
...
+	enrichMcpRequestProps(request, _ctx, baseProps): Props {
```
Rifdhan, line 123: This function name feels pretty generic, it looks like it is only adding API version so maybe we should call it addApiVersionToProps?

src/index.ts
```diff
@@ -36,78 +44,132 @@ const config: ResolveConfigFn = (env: Env, _trigger) => {
...
+				: undefined,
+			apiVersionMode,
+		};
+	},
+	// Extra routes mounted on the default handler app (consumer-specific).
+	extraRoutes(app) {
+		app.get("/", async (c) => {
+			return c.env.ASSETS!.fetch("/index.html");
+		});
+		app.get("/hello", (c) => c.json({ message: "Hello, World!" }));
```
Rifdhan: Let's just get rid of this /hello endpoint

src/index.ts
adityamittal3107: addApiVersionToProps would under-describe it. The generic name is actually intentional — the package documents it as a general-purpose "inject props" extension point ("apply per-request prop injection outside the bearer flow"), and consumers are free to enrich props however they like.

src/index.ts
adityamittal3107: done

src/utils.ts
```diff
@@ -15,83 +19,52 @@ export type Props = {
...
+): Props["clientName"] {
+	return {
+		clientId: clientName?.clientId ?? DEFAULT_CLIENT_NAME,
+		clientName: clientName?.clientName ?? DEFAULT_CLIENT_NAME,
+		registrationDate:
+			(clientName && "registrationDate" in clientName
+				? clientName.registrationDate
+				: undefined) ?? Date.now(),
+	};
+}
```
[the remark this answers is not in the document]
adityamittal3107: The type is ClientInfo | { clientId; clientName; registrationDate } | null never a string.

src/index.ts
adityamittal3107: Yeah, we have that in auth package. Remove any

src/index.ts
adityamittal3107: Yes done.

Rifdhan (approved):

PR #176, opened 2026-07-07
SCAL-319970: Multi-org support (list_orgs / switch_org, keep-warm token, idle TTL)

rohitthughtspot:
## Summary

- **`list_orgs` / `switch_org` tools**: new MCP tools gated on OAuth + orgs-enabled cluster + v2 API surface + a post-multi-org grant (refresh token present). Org-scoped bearer tokens are minted via `/callosum/v1/v2/auth/token/fetch` and persisted in durable storage so concurrent sessions share one mint.
- **Keep-warm token** (`UserTokenStoreSQLite` Durable Object): a DO alarm fires every 11h to call `gettoken?refresh=true`, preventing the 24h token expiry from killing long-idle sessions. Idle TTL of 14 days abandons storage for users who haven't connected in two weeks.
- **`loadOrSeedWarmToken`**: read-before-write on connect — if the alarm-refreshed token is newer than the props token (different string, not expired), use it directly; otherwise seed from the fresh OAuth grant.
- **Backward compat**: grants minted before multi-org shipped have no refresh token → org tools stay hidden until re-auth. No code change needed for existing users.
- **`resources: { listChanged: true }` capability**: `switch_org` sends a `notifications/resources/list_changed` notification so clients re-fetch the new org's datasources.
- **`@thoughtspot/mcp-auth` bumped to `^2.0.0`**: picks up the `refreshToken` field added to the OAuth grant.
- **`is_active` omitted for inactive orgs**: response only includes `is_active: true` on the active org, saving tokens at scale.
- **Wrong-org hint in `get_session_updates`**: when no data is found, the LLM is told the data may be in a different org or the user may not have access, and to suggest `list_orgs` / `switch_org`.
- **Tests**: 628 passing — new suites cover `UserTokenStoreSQLite` (keep-warm alarm, idle TTL, lastSeenAt), `MCPServer` org flows (list, switch, token retry, backward compat), and `OrgService`.

## Test plan

- [ ] Connect with an OAuth session — `list_orgs` appears in tool list on an orgs-enabled cluster
- [ ] Call `list_orgs` — returns org list; only active org has `is_active: true`, others have no `is_active` field
- [ ] Call `switch_org` with a valid org ID — subsequent data calls use org-scoped token
- [ ] Call `switch_org` with an invalid org ID — returns error, active org unchanged
- [ ] Connect with a legacy grant (no refresh token) — `list_orgs`/`switch_org` absent from tool list
- [ ] Verify keep-warm alarm fires at ~11h intervals in wrangler tail
- [ ] `npm test` — 628 passed

### DIFF
--- a/src/servers/mcp-server-base.ts
+++ b/src/servers/mcp-server-base.ts
@@ -28,6 +28,7 @@ import {
 } from "../metrics/runtime/tool-metrics";
 import { getActiveSpan, withSpan } from "../metrics/tracing/tracing-utils";
 import { StorageServiceClient } from "../storage-service/storage-service";
+import { OrgService } from "../thoughtspot/org-service";
 import { getThoughtSpotClient } from "../thoughtspot/thoughtspot-client";
 import { ThoughtSpotService } from "../thoughtspot/thoughtspot-service";
 import type { Props } from "../utils";
@@ -91,6 +92,18 @@ export abstract class BaseMCPServer extends Server {
 			return false;
 		}
 		return String(this.sessionInfo.enableSpotterDataSourceDiscovery) === "true";
+	}
+
+	/**
+	 * Whether Orgs are enabled on this cluster (from session info). Fails closed:
+	 * if session info is unavailable or the flag is absent, returns false so the
+	 * org tools stay hidden.
+	 */
+	protected isOrgsEnabled(): boolean {
+		if (!this.sessionInfo) {
+			return false;
+		}
+		return this.sessionInfo.orgsEnabled === true;
 	}
 
 	/**
@@ -187,23 +200,60 @@ export abstract class BaseMCPServer extends Server {
 		};
 	}
 
-	protected async getStorageService(): Promise<StorageServiceClient> {
-		const accessToken = this.ctx.props.accessToken;
-		if (!accessToken || accessToken.length === 0) {
-			throw new Error("Access token is required to use Storage Service");
+	/**
+	 * Stable per-login hash used to namespace this user's durable storage (both
+	 * conversation buffers and active-org state), keeping users isolated.
+	 *
+	 * Keyed on the refresh token when present (OAuth): it is stable across the
+	 * access token's 24h rotation and only changes on full reauthentication, so
+	 * storage survives token refresh and resets on reauth. Falls back to the
+	 * access token for static bearer/token connections, which have no refresh
+	 * token (their token is long-lived).
+	 */
+	protected async getStorageKeyHash(): Promise<string> {
+		const keyToken = this.ctx.props.refreshToken ?? this.ctx.props.accessToken;
+		if (!keyToken || keyToken.length === 0) {
+			throw new Error("A token is required to derive the storage key");
 		}
-		const encodedAccessToken = new TextEncoder().encode(accessToken);
 		const hashBuffer = await crypto.subtle.digest(
 			"SHA-256",
-			encodedAccessToken,
+			new TextEncoder().encode(keyToken),
 		);
-		const hashUrlSafe = Buffer.from(new Uint8Array(hashBuffer)).toString(
-			"base64url",
-		);
+		return Buffer.from(new Uint8Array(hashBuffer)).toString("base64url");
+	}
+
+	protected async getStorageService(): Promise<StorageServiceClient> {
+		const hashUrlSafe = await this.getStorageKeyHash();
 		return new StorageServiceClient(
 			this.ctx.env
 				.CONVERSATION_STORAGE_OBJECT as unknown as DurableObjectNamespace,
 			hashUrlSafe,
+			this.ctx.env.USER_TOKEN_OBJECT as unknown as DurableObjectNamespace,
+		);
+	}
+
+	// Active org for ThoughtSpot calls (x-thoughtspot-orgs header); none by default.
+	// Subclasses override to supply per-session org state.
+	protected getActiveOrgId(): string | undefined {
+		return undefined;
+	}
+
+	// Bearer token for ThoughtSpot calls; the session token by default. Subclasses
+	// override to return an org-scoped token.
+	protected getActiveBearerToken(): string {
+		return this.ctx.props.accessToken;
+	}
+
+	// Build an OrgService bound to an explicit (cluster-wide) token for org listing
+	// / org-token minting.
+	protected getOrgService(
+		bearerToken: string,
+		orgId?: string,
+		recorder?: MetricsRecorder,
+	) {
+		return new OrgService(
+			getThoughtSpotClient(this.ctx.props.instanceUrl, bearerToken, orgId),
+			recorder,
 		);
 	}
 
@@ -214,7 +264,8 @@ export abstract class BaseMCPServer extends Server {
 		return new ThoughtSpotService(
 			getThoughtSpotClient(
 				this.ctx.props.instanceUrl,
-				this.ctx.props.accessToken,
+				this.getActiveBearerToken(),
+				this.getActiveOrgId(),
 			),
 			{
 				recorder,
@@ -439,7 +490,21 @@ export abstract class BaseMCPServer extends Server {
 				});
 			},
 		);
+
+		// Subclass post-initialization hook (runs after sessionInfo is available
+		// and handlers are registered). Best-effort: failures must not break the
+		// connection.
+		try {
+			await this.postInit();
+		} catch (error) {
+			console.error("postInit failed:", error);
+		}
 	}
+
+	/**
+	 * Optional hook for subclasses to run setup after init(). Default no-op.
+	 */
+	protected async postInit(): Promise<void> {}
 
 	async addTracker(tracker: Tracker) {
 		this.trackers.add(tracker);
--- a/src/servers/mcp-server-base.ts
+++ b/src/servers/mcp-server-base.ts
@@ -234,17 +234,8 @@ export abstract class BaseMCPServer extends Server {
 		);
 	}
 
-	// Active org for ThoughtSpot calls (x-thoughtspot-orgs header); none by default.
-	// Subclasses override to supply per-session org state.
-	protected getActiveOrgId(): string | undefined {
-		return undefined;
-	}
-
-	// Bearer token for ThoughtSpot calls; the session token by default. Subclasses
-	// override to return an org-scoped token.
-	protected getActiveBearerToken(): string {
-		return this.ctx.props.accessToken;
-	}
+	protected abstract getActiveOrgId(): string | undefined;
+	protected abstract getActiveOrgToken(): string;
 
 	// Build an OrgService bound to an explicit (cluster-wide) token for org listing
 	// / org-token minting.
@@ -266,7 +257,7 @@ export abstract class BaseMCPServer extends Server {
 		return new ThoughtSpotService(
 			getThoughtSpotClient(
 				this.ctx.props.instanceUrl,
-				this.getActiveBearerToken(),
+				this.getActiveOrgToken(),
 				this.getActiveOrgId(),
 			),
 			{
--- a/src/servers/mcp-server-base.ts
+++ b/src/servers/mcp-server-base.ts
@@ -235,7 +235,9 @@ export abstract class BaseMCPServer extends Server {
 	}
 
 	protected abstract getActiveOrgId(): string | undefined;
-	protected abstract getActiveOrgToken(): string;
+	// The token to authenticate the current request: the org-scoped token when an
+	// org is active, otherwise the global/cluster token.
+	protected abstract getActiveToken(): string;
 
 	// Build an OrgService bound to an explicit (cluster-wide) token for org listing
 	// / org-token minting.
@@ -257,7 +259,7 @@ export abstract class BaseMCPServer extends Server {
 		return new ThoughtSpotService(
 			getThoughtSpotClient(
 				this.ctx.props.instanceUrl,
-				this.getActiveOrgToken(),
+				this.getActiveToken(),
 				this.getActiveOrgId(),
 			),
 			{
--- a/src/servers/mcp-server-base.ts
+++ b/src/servers/mcp-server-base.ts
@@ -75,7 +75,9 @@ export abstract class BaseMCPServer extends Server {
 				capabilities: {
 					tools: {},
 					completion: {},
-					resources: {},
+					// listChanged: the resource set (datasources) changes on switch_org,
+					// so we notify clients to re-list.
+					resources: { listChanged: true },
 				},
 			},
 		);
--- a/src/servers/mcp-server-base.ts
+++ b/src/servers/mcp-server-base.ts
@@ -27,6 +27,7 @@ import {
 	recordToolInvocationMetrics,
 } from "../metrics/runtime/tool-metrics";
 import { getActiveSpan, withSpan } from "../metrics/tracing/tracing-utils";
+import { OrgStorageServiceClient } from "../storage-service/org-storage-service";
 import { StorageServiceClient } from "../storage-service/storage-service";
 import { OrgService } from "../thoughtspot/org-service";
 import { getThoughtSpotClient } from "../thoughtspot/thoughtspot-client";
@@ -213,7 +214,8 @@ export abstract class BaseMCPServer extends Server {
 	 * token (their token is long-lived).
 	 */
 	protected async getStorageKeyHash(): Promise<string> {
-		const keyToken = this.ctx.props.refreshToken ?? this.ctx.props.accessToken;
+		const keyToken =
+			this.ctx.props.globalRefreshToken ?? this.ctx.props.accessToken;
 		if (!keyToken || keyToken.length === 0) {
 			throw new Error("A token is required to derive the storage key");
 		}
@@ -230,7 +232,14 @@ export abstract class BaseMCPServer extends Server {
 			this.ctx.env
 				.CONVERSATION_STORAGE_OBJECT as unknown as DurableObjectNamespace,
 			hashUrlSafe,
+		);
+	}
+
+	protected async getOrgStorageService(): Promise<OrgStorageServiceClient> {
+		const hashUrlSafe = await this.getStorageKeyHash();
+		return new OrgStorageServiceClient(
 			this.ctx.env.USER_TOKEN_OBJECT as unknown as DurableObjectNamespace,
+			hashUrlSafe,
 		);
 	}
 
--- a/src/servers/mcp-server.ts
+++ b/src/servers/mcp-server.ts
@@ -11,10 +11,18 @@ import {
 	type MetricsRecorder,
 	NOOP_METRICS_RECORDER,
 } from "../metrics/runtime/metrics-recorder";
+import type { MetricAnalyticsContext } from "../metrics/runtime/metrics-sink";
 import type { ToolMetricApiSurface } from "../metrics/runtime/tool-metrics";
 import { WithSpan } from "../metrics/tracing/tracing-utils";
-import type { DataSource } from "../thoughtspot/thoughtspot-service";
-import type { Answer, StreamingMessagesState } from "../thoughtspot/types";
+import type {
+	DataSource,
+	ThoughtSpotService,
+} from "../thoughtspot/thoughtspot-service";
+import {
+	type Answer,
+	type StreamingMessagesState,
+	ThoughtSpotApiError,
+} from "../thoughtspot/types";
 import { McpServerError } from "../utils";
 import { BaseMCPServer, type Context } from "./mcp-server-base";
 import {
@@ -26,6 +34,7 @@ import {
 	GetRelevantQuestionsSchema,
 	GetSessionUpdatesInputSchema,
 	SendSessionMessageInputSchema,
+	SwitchOrgInputSchema,
 	ToolName,
 } from "./tool-definitions";
 import {
@@ -35,12 +44,268 @@ import {
 } from "./version-registry";
 
 export class MCPServer extends BaseMCPServer {
+	// Mirrors of the shared per-user store (the durable source of truth, shared so
+	// the token is minted once across the user's fanned-out sessions).
+	private activeOrgId: string | undefined;
+	private activeOrgToken: string | undefined;
+	private warmGlobalToken: string | undefined;
+
 	constructor(ctx: Context) {
 		super(ctx, "ThoughtSpot", "2.0.0");
 	}
 
+	private getGlobalToken(): string {
+		return this.warmGlobalToken ?? this.ctx.props.accessToken;
+	}
+
+	protected getActiveOrgId(): string | undefined {
+		return this.activeOrgId;
+	}
+
+	protected getActiveBearerToken(): string {
+		// An active org must use its org-scoped token — never the global token, which
+		// would break org isolation. ensureActiveOrgToken guarantees it is minted
+		// before any data call, so a missing token here is a bug, not a fallback.
+		if (this.activeOrgId) {
+			if (!this.activeOrgToken) {
+				throw new Error(
+					`Org ${this.activeOrgId} is active but its token is not minted`,
+				);
+			}
+			return this.activeOrgToken;
+		}
+		return this.getGlobalToken();
+	}
+
+	// Re-read on each org-aware call so a switch in another session is reflected.
+	private async loadActiveOrg(): Promise<void> {
+		const storage = await this.getStorageService();
+		const stored = await storage.getActiveOrg();
+		this.activeOrgId = stored.activeOrgId ?? undefined;
+		this.activeOrgToken = stored.orgToken ?? undefined;
+	}
+
+	// Pass orgToken to commit id+token atomically (validated switch); else the
+	// prior token is cleared and re-minted lazily.
+	private async setActiveOrg(orgId: string, orgToken?: string): Promise<void> {
+		this.activeOrgId = orgId;
+		this.activeOrgToken = orgToken;
+		const storage = await this.getStorageService();
+		await storage.setActiveOrg(orgId, orgToken);
+	}
+
+	// Reuse the shared-store token if held, else mint from the global token and
+	// persist it (so the fan-out mints once).
+	private async ensureOrgToken(
+		orgId: string,
+		recorder?: MetricsRecorder,
+	): Promise<string> {
+		if (this.activeOrgId === orgId && this.activeOrgToken) {
+			return this.activeOrgToken;
+		}
+		await this.loadOrSeedWarmToken();
+		const globalToken = this.getGlobalToken();
+		const orgToken = await this.getOrgService(
+			globalToken,
+			undefined,
+			recorder,
+		).fetchOrgBearerToken(globalToken, orgId);
+		this.activeOrgToken = orgToken;
+		const storage = await this.getStorageService();
+		await storage.setActiveOrgToken(orgToken);
+		return orgToken;
+	}
+
+	// HTTP status from a thrown error or `{ error }` result: typed
+	// ThoughtSpotApiError, else parsed from a "status NNN" message.
+	private apiErrorStatus(value: unknown): number | undefined {
+		const err =
+			value instanceof Error
+				? value
+				: ((value as { error?: unknown } | null)?.error ?? value);
+		if (err instanceof ThoughtSpotApiError) {
+			return err.status;
+		}
+		const message =
+			typeof (err as { message?: unknown })?.message === "string"
+				? (err as { message: string }).message
+				: "";
+		const match = message.match(/\bstatus (\d{3})\b/);
+		return match ? Number(match[1]) : undefined;
+	}
+
+	// Re-mint the active org's token, recovering from a stale one. Mint first, then
+	// overwrite atomically — the store never goes tokenless, so a concurrent session
+	// never sees an active org without its token. Abort if a sibling switched org
+	// meanwhile, so we don't clobber the new org's token.
+	private async forceRemintOrgToken(
+		orgId: string,
+		recorder?: MetricsRecorder,
+	): Promise<void> {
+		await this.loadActiveOrg();
+		if (this.activeOrgId !== orgId) {
+			return;
+		}
+		await this.loadOrSeedWarmToken();
+		const globalToken = this.getGlobalToken();
+		const orgToken = await this.getOrgService(
+			globalToken,
+			undefined,
+			recorder,
+		).fetchOrgBearerToken(globalToken, orgId);
+		this.activeOrgToken = orgToken;
+		await this.setActiveOrg(orgId, orgToken);
+	}
+
+	// Single re-mint+retry on a 401, only when an org token is in use (a global-token
+	// 401 passes through). Handles both thrown and `{ error }`-result 401s.
+	protected async withOrgTokenRetry<T>(
+		recorder: MetricsRecorder | undefined,
+		fn: (service: ThoughtSpotService) => Promise<T>,
+		analyticsContextOverride?: MetricAnalyticsContext,
+	): Promise<T> {
+		const orgId = this.activeOrgId;
+
+		const attempt = () =>
+			fn(this.getThoughtSpotService(recorder, analyticsContextOverride));
+
+		if (!orgId) {
+			return attempt();
+		}
+
+		// Org active → the call must use the org-scoped token, never the global one
+		// (a global-token data call would break org isolation). Mint if not cached.
+		if (!this.activeOrgToken) {
+			await this.ensureOrgToken(orgId, recorder);
+		}
+
+		try {
+			const result = await attempt();
+			if (this.apiErrorStatus(result) === 401) {
+				await this.forceRemintOrgToken(orgId, recorder);
+				return attempt();
+			}
+			return result;
+		} catch (error) {
+			if (this.apiErrorStatus(error) === 401) {
+				await this.forceRemintOrgToken(orgId, recorder);
+				return attempt();
+			}
+			throw error;
+		}
+	}
+
+	// validateConnection maps a 401 to `false` (no throw), so withOrgTokenRetry can't
+	// see it. If it fails with an org token in use, re-mint and re-validate.
+	private async validateConnectionWithOrgRetry(
+		recorder?: MetricsRecorder,
+	): Promise<boolean> {
+		const orgId = this.activeOrgId;
+		const usedOrgToken = Boolean(orgId && this.activeOrgToken);
+		const ok = await this.getThoughtSpotService(recorder).validateConnection();
+		if (ok || !usedOrgToken || !orgId) {
+			return ok;
+		}
+		await this.forceRemintOrgToken(orgId, recorder);
+		return this.getThoughtSpotService(recorder).validateConnection();
+	}
+
+	// On connect (OAuth only): keep the cluster token warm, then — when org tools are
+	// available — establish the active org (prior switch wins, else session current)
+	// and mint its token. Best-effort: never break connect.
+	protected async postInit(): Promise<void> {
+		if (!this.isOAuthAuth()) {
+			return;
+		}
+		try {
+			await this.loadOrSeedWarmToken();
+		} catch (error) {
+			console.error("Failed to load/seed keep-warm token on connect:", error);
+		}
+
+		if (!this.areOrgToolsAvailable()) {
+			return;
+		}
+		try {
+			await this.loadActiveOrg();
+			if (!this.activeOrgId) {
+				const currentOrgId =
+					this.sessionInfo?.currentOrgId != null
+						? String(this.sessionInfo.currentOrgId)
+						: undefined;
+				if (currentOrgId) {
+					await this.setActiveOrg(currentOrgId);
+				}
+			}
+			if (this.activeOrgId) {
+				await this.ensureOrgToken(this.activeOrgId);
+			}
+		} catch (error) {
+			console.error("Failed to set/mint active org on connect:", error);
+		}
+	}
+
+	// Load the keep-warm token into memory. Re-seed from props' fresh grant when the
+	// store is unseeded or expired (the refresh chain died); else the store wins.
+	private async loadOrSeedWarmToken(): Promise<void> {
+		const storage = await this.getStorageService();
+		const store = await storage.getTokenStore();
+		const storedExpired =
+			typeof store.expiresAt === "number" && store.expiresAt <= Date.now();
+		if (store.accessToken && !storedExpired) {
+			this.warmGlobalToken = store.accessToken;
+			return;
+		}
+		const { accessToken, refreshToken, tokenExpiryDuration, instanceUrl } =
+			this.ctx.props;
+		if (accessToken && refreshToken) {
+			await storage.seedTokenStore({
+				accessToken,
+				refreshToken,
+				instanceUrl,
+				expiresAt:
+					typeof tokenExpiryDuration === "number"
+						? tokenExpiryDuration
+						: undefined,
+			});
+			this.warmGlobalToken = accessToken;
+			return;
+		}
+		// No refresh token to re-seed with: use whatever we have.
+		this.warmGlobalToken = store.accessToken ?? accessToken;
+	}
+
+	// Fire-and-forget; throttle + idle-delete live in the DO.
+	private touchLastSeen(): void {
+		this.getStorageService()
+			.then((storage) => storage.touchLastSeen())
+			.catch((error) => {
+				console.error("Failed to record last-seen activity:", error);
+			});
+	}
+
 	protected getToolMetricApiSurface(): ToolMetricApiSurface {
 		return "mcp";
+	}
+
+	// Gates OAuth-only tools such as `list_orgs`.
+	protected isOAuthAuth(): boolean {
+		return this.ctx.props.authMode === "oauth";
+	}
+
+	// Org behavior requires OAuth + orgs enabled + the v2 surface (inferred from the
+	// resolved tool set, not a label). Fails closed.
+	protected areOrgToolsAvailable(): boolean {
+		if (!this.isOAuthAuth() || !this.isOrgsEnabled()) {
+			return false;
+		}
+		try {
+			return resolveApiVersion(this.ctx.props.apiVersion).tools.some(
+				(tool) => tool?.name === ToolName.ListOrgs,
+			);
+		} catch {
+			return false;
+		}
 	}
 
 	protected getToolMetricApiVersionLabel(): string | undefined {
@@ -136,6 +401,13 @@ export class MCPServer extends BaseMCPServer {
 			);
 		}
 
+		if (!this.areOrgToolsAvailable()) {
+			tools = tools.filter(
+				(tool) =>
+					tool.name !== ToolName.ListOrgs && tool.name !== ToolName.SwitchOrg,
+			);
+		}
+
 		return { tools };
 	}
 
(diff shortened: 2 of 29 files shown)


src/servers/mcp-server.ts
```diff
@@ -612,19 +961,110 @@ Provide this url to the user as a link to view the liveboard in ThoughtSpot.`;
...
+		span?.setAttribute("active_org_id", orgId);
+
+		// Datasources are org-specific — tell the client to re-list resources so it
+		// drops the previous org's datasources. Best-effort: never fail the switch.
+		try {
+			await this.sendResourceListChanged();
+			// TEMP (testing): confirm the notification path fires on switch. Remove.
+			console.log(
+				`[RESOURCE-DEBUG] sendResourceListChanged sent after switch to org ${orgId}`,
+			);
```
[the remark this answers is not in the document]
rohitthughtspot: done

src/servers/mcp-server-base.ts
```diff
@@ -187,23 +202,60 @@ export abstract class BaseMCPServer extends Server {
...
+
+	// Active org for ThoughtSpot calls (x-thoughtspot-orgs header); none by default.
+	// Subclasses override to supply per-session org state.
+	protected getActiveOrgId(): string | undefined {
+		return undefined;
+	}
+
+	// Bearer token for ThoughtSpot calls; the session token by default. Subclasses
+	// override to return an org-scoped token.
+	protected getActiveBearerToken(): string {
```
Rifdhan: getActiveOrgToken

src/servers/mcp-server-base.ts
```diff
@@ -187,23 +202,60 @@ export abstract class BaseMCPServer extends Server {
...
 				.CONVERSATION_STORAGE_OBJECT as unknown as DurableObjectNamespace,
 			hashUrlSafe,
+			this.ctx.env.USER_TOKEN_OBJECT as unknown as DurableObjectNamespace,
+		);
+	}
+
+	// Active org for ThoughtSpot calls (x-thoughtspot-orgs header); none by default.
+	// Subclasses override to supply per-session org state.
+	protected getActiveOrgId(): string | undefined {
+		return undefined;
```
Rifdhan: Why do we have a dummy implementation instead of just a stub like we have for postInit?

src/servers/mcp-server.ts
```diff
@@ -35,14 +44,258 @@ import {
 } from "./version-registry";
 
 export class MCPServer extends BaseMCPServer {
+	private activeOrgId: string | undefined;
+	private activeOrgToken: string | undefined;
+	private warmGlobalToken: string | undefined;
```
Rifdhan: I don't think "warm" is needed, let's just call it globalToken

src/servers/mcp-server.ts
```diff
@@ -35,14 +44,258 @@ import {
...
+	// False for pre-multi-org grants (no refresh token) — keeps old behavior until re-auth.
+	private grantHasRefreshToken = false;
+
 	constructor(ctx: Context) {
 		super(ctx, "ThoughtSpot", "2.0.0");
 	}
 
+	private getGlobalToken(): string {
+		return this.warmGlobalToken ?? this.ctx.props.accessToken;
```
Rifdhan: accessToken is not a global token, either the function name should be different or this is incorrect logic?

src/servers/mcp-server.ts
```diff
@@ -35,14 +44,258 @@ import {
...
+	}
+
+	protected getActiveOrgId(): string | undefined {
```
Rifdhan, line 56: Nit: no need for a wrapper function, just use the private property

src/servers/mcp-server.ts
```diff
@@ -35,14 +44,258 @@ import {
...
+		return this.activeOrgId;
+	}
+
+	protected getActiveBearerToken(): string {
```
Rifdhan: getActiveOrgToken

src/servers/mcp-server.ts
```diff
@@ -35,14 +44,258 @@ import {
...
+		if (this.activeOrgId) {
+			if (!this.activeOrgToken) {
+				throw new Error(
+					`Org ${this.activeOrgId} is active but its token is not minted`,
+				);
+			}
+			return this.activeOrgToken;
+		}
+		return this.getGlobalToken();
```
Rifdhan: Similarly, the global token is not the active org token and cannot be used in place of that, so this seems wrong

src/servers/mcp-server.ts
```diff
@@ -35,14 +44,258 @@ import {
...
+	}
+
+	private async setActiveOrg(orgId: string, orgToken?: string): Promise<void> {
+		this.activeOrgId = orgId;
+		this.activeOrgToken = orgToken;
+		const storage = await this.getStorageService();
+		await storage.setActiveOrg(orgId, orgToken);
+	}
+
+	private async ensureOrgToken(
```
Rifdhan: getOrRecreateActiveOrgToken

src/servers/mcp-server.ts
```diff
@@ -35,14 +44,258 @@ import {
...
+			return this.activeOrgToken;
+		}
+		await this.loadOrSeedWarmToken();
+		const globalToken = this.getGlobalToken();
+		const orgToken = await this.getOrgService(
+			globalToken,
+			undefined,
+			recorder,
+		).fetchOrgBearerToken(globalToken, orgId);
+		this.activeOrgToken = orgToken;
```
Rifdhan: Should we call setActiveOrg for this last part? Duplication of logic

src/servers/mcp-server.ts
```diff
@@ -35,14 +44,258 @@ import {
...
+		}
+		const message =
+			typeof (err as { message?: unknown })?.message === "string"
+				? (err as { message: string }).message
+				: "";
+		const match = message.match(/\bstatus (\d{3})\b/);
+		return match ? Number(match[1]) : undefined;
+	}
+
+	private async forceRemintOrgToken(
```
Rifdhan: forceRecreateActiveOrgToken

src/servers/mcp-server.ts
```diff
@@ -35,14 +44,258 @@ import {
...
+		orgId: string,
+		recorder?: MetricsRecorder,
+	): Promise<void> {
+		await this.loadActiveOrg();
+		if (this.activeOrgId !== orgId) {
+			return;
```
Rifdhan, line 126: The method should use this.activeOrgId directly and the caller should just not call this if it's not the org they want

src/servers/mcp-server.ts
```diff
@@ -35,14 +44,258 @@ import {
...
+		const orgToken = await this.getOrgService(
+			globalToken,
+			undefined,
+			recorder,
+		).fetchOrgBearerToken(globalToken, orgId);
+		this.activeOrgToken = orgToken;
+		await this.setActiveOrg(orgId, orgToken);
+	}
+
+	protected async withOrgTokenRetry<T>(
```
Rifdhan: I want to get rid of org token retry, since we should not need it anymore. If we refresh the org token every 11h then it should never need to be retried due to expiry?

src/servers/mcp-server.ts
```diff
@@ -35,14 +44,258 @@ import {
...
+
 	protected getToolMetricApiSurface(): ToolMetricApiSurface {
 		return "mcp";
 	}
 
+	protected isOAuthAuth(): boolean {
+		return this.ctx.props.authMode === "oauth";
+	}
+
+	protected hasMultiOrgGrant(): boolean {
```
Rifdhan: Ditto no need for wrapper function

src/servers/mcp-server.ts
```diff
@@ -35,14 +44,258 @@ import {
...
+	protected areOrgToolsAvailable(): boolean {
+		if (
+			!this.isOAuthAuth() ||
+			!this.isOrgsEnabled() ||
+			!this.hasMultiOrgGrant()
+		) {
+			return false;
+		}
+		try {
+			return resolveApiVersion(this.ctx.props.apiVersion).tools.some(
```
Rifdhan, line 270: API version gating should be handled inside listTools instead of in here

src/servers/mcp-server.ts
```diff
@@ -189,10 +449,26 @@ export class MCPServer extends BaseMCPServer {
...
+
+		if (this.areOrgToolsAvailable()) {
+			await this.loadActiveOrg();
+			if (
+				this.activeOrgId &&
+				!this.activeOrgToken &&
+				name !== ToolName.ListOrgs &&
+				name !== ToolName.SwitchOrg
+			) {
+				await this.ensureOrgToken(this.activeOrgId, recorder);
```
Rifdhan: We don't need to do this here, the org token should be valid since we refresh it every 11h

src/servers/mcp-server.ts
```diff
@@ -612,19 +909,101 @@ Provide this url to the user as a link to view the liveboard in ThoughtSpot.`;
...
+		let orgToken: string;
+		try {
+			await this.loadOrSeedWarmToken();
+			const globalToken = this.getGlobalToken();
+			orgToken = await this.getOrgService(
+				globalToken,
+				undefined,
+				recorder,
+			).fetchOrgBearerToken(globalToken, orgId);
+		} catch (error) {
```
Rifdhan, line 975: Let's console.error the error as well

src/servers/user-token-store-server.ts
```diff
@@ -0,0 +1,213 @@
...
+					if (body.orgToken) {
+						await this.state.storage.put({
+							[ACTIVE_ORG_KEY]: body.activeOrgId,
+							[ORG_TOKEN_KEY]: body.orgToken,
+						});
+						return Response.json({ ok: true });
+					}
+					const previousOrgId =
+						await this.state.storage.get<string>(ACTIVE_ORG_KEY);
+					await this.state.storage.put<string>(
```
Rifdhan: Why not put this in the if-block below too? No need to store if it's already the same

src/servers/user-token-store-server.ts
```diff
@@ -0,0 +1,213 @@
...
+	) {}
+
+	async fetch(request: Request): Promise<Response> {
+		const url = new URL(request.url);
+		const parts = url.pathname.split("/");
+		const operation = parts[3] ?? "";
+
+		try {
+			switch (`${request.method} /${operation}`) {
+				case "GET /active-org": {
```
Rifdhan: GET /active-org-id-and-token

src/servers/user-token-store-server.ts
```diff
@@ -0,0 +1,213 @@
...
+					]);
+					return Response.json({
+						activeOrgId: activeOrgId ?? null,
+						orgToken: orgToken ?? null,
+					});
+				}
+
+				// Clear the token only on a real org change; re-setting the same org
+				// (every cold connect) must not delete a token a sibling just minted.
+				case "POST /active-org": {
```
Rifdhan: POST /active-org-id-and-token

src/servers/user-token-store-server.ts
```diff
@@ -0,0 +1,213 @@
...
+					const body = (await request.json()) as { orgToken?: string | null };
+					if (body.orgToken) {
+						await this.state.storage.put<string>(ORG_TOKEN_KEY, body.orgToken);
+					} else {
+						await this.state.storage.delete(ORG_TOKEN_KEY);
+					}
+					return Response.json({ ok: true });
+				}
+
+				case "GET /token-store": {
```
Rifdhan: GET /global-token-data

src/servers/user-token-store-server.ts
```diff
@@ -0,0 +1,213 @@
...
+					const store =
+						(await this.state.storage.get<TokenStore>(TOKEN_STORE_KEY)) ?? null;
+					return Response.json({
+						accessToken: store?.accessToken ?? null,
+						expiresAt: store?.expiresAt ?? null,
+					});
+				}
+
+				case "POST /token-store": {
```
Rifdhan: POST /global-token-data

src/servers/user-token-store-server.ts
```diff
@@ -0,0 +1,213 @@
+const ACTIVE_ORG_KEY = "active-org";
```
Rifdhan: active-org-id

src/servers/user-token-store-server.ts
```diff
@@ -0,0 +1,213 @@
...
+const ORG_TOKEN_KEY = "active-org-token";
+const TOKEN_STORE_KEY = "token-store";
```
Rifdhan: global-token-details

src/servers/user-token-store-server.ts
```diff
@@ -0,0 +1,213 @@
...
+const TOKEN_REFRESH_INTERVAL_MS = 11 * 60 * 60 * 1000;
```
Rifdhan: GLOBAL_TOKEN_REFRESH_INTERVAL_MS

src/servers/user-token-store-server.ts
```diff
@@ -0,0 +1,213 @@
...
+const SESSION_IDLE_TTL_MS = 14 * 24 * 60 * 60 * 1000;
+
+export type TokenStore = {
+	accessToken: string;
```
Rifdhan: globalToken

src/servers/user-token-store-server.ts
```diff
@@ -0,0 +1,213 @@
...
+	refreshToken: string;
```
Rifdhan: globalRefreshToken

src/servers/user-token-store-server.ts
```diff
@@ -0,0 +1,213 @@
...
+					});
+				}
+
+				case "POST /token-store": {
+					const body = (await request.json()) as TokenStore;
+					await this.seedTokenStore(body);
+					return Response.json({ ok: true });
+				}
+
+				case "POST /touch": {
```
Rifdhan: POST /last-seen

src/servers/user-token-store-server.ts
```diff
@@ -0,0 +1,213 @@
+const ACTIVE_ORG_KEY = "active-org";
+const ORG_TOKEN_KEY = "active-org-token";
+const TOKEN_STORE_KEY = "token-store";
+const TOKEN_REFRESH_INTERVAL_MS = 11 * 60 * 60 * 1000;
+const SESSION_IDLE_TTL_MS = 14 * 24 * 60 * 60 * 1000;
+
+export type TokenStore = {
```
Rifdhan: GlobalTokenData

rohitthughtspot: done

src/servers/user-token-store-server.ts
rohitthughtspot: done

src/servers/user-token-store-server.ts
rohitthughtspot: done

src/servers/user-token-store-server.ts
rohitthughtspot: done

src/servers/user-token-store-server.ts
rohitthughtspot: done

src/servers/user-token-store-server.ts
rohitthughtspot: done

src/servers/user-token-store-server.ts
rohitthughtspot: done

src/servers/user-token-store-server.ts
rohitthughtspot: done

src/servers/user-token-store-server.ts
rohitthughtspot: done

src/servers/user-token-store-server.ts
rohitthughtspot: done

src/servers/user-token-store-server.ts
rohitthughtspot: done

src/servers/user-token-store-server.ts
rohitthughtspot: done

src/servers/mcp-server.ts
rohitthughtspot: done

src/servers/mcp-server.ts
rohitthughtspot: removed it

src/servers/mcp-server.ts
rohitthughtspot: done

src/servers/mcp-server.ts
rohitthughtspot: done

src/servers/mcp-server.ts
rohitthughtspot: done

src/servers/mcp-server.ts
rohitthughtspot: removed

src/servers/mcp-server.ts
rohitthughtspot: removed it

src/servers/mcp-server.ts
rohitthughtspot: done

src/servers/mcp-server.ts
rohitthughtspot: done

src/servers/mcp-server.ts
rohitthughtspot: done

src/servers/mcp-server-base.ts
rohitthughtspot: done

src/servers/mcp-server-base.ts
rohitthughtspot: done

src/servers/mcp-server.ts
rohitthughtspot: can do this as part of seperate pr as this will create duplicate code we need this for list/switch orgs depending on version

src/servers/mcp-server.ts
rohitthughtspot: done

src/servers/mcp-server.ts
rohitthughtspot: renamed as this will be needed in cases of org disabled

static/oauth-callback.js
```diff
@@ -130,7 +140,23 @@
...
+        // gettoken returns the access token + refreshToken + timestamps. Normalize
+        // to the { data: { token, ... } } shape /store-token expects.
+        const src = (raw && raw.data) ? raw.data : raw;
+        if (!src || !src.token) {
+            // Fail loudly at login rather than storing an undefined token and
+            // failing opaquely on the first tool call.
+            throw new Error('Authentication response did not contain a token.');
+        }
+        const data = {
+            data: {
```
Rifdhan, line 141: Let's follow our naming convention - globalToken, globalRefreshToken, globalTokenCreatedAt, globalTokenExpiryDuration

src/servers/mcp-server.ts
```diff
@@ -35,14 +43,189 @@ import {
...
+		const orgToken = await this.getOrgService(
+			globalToken,
+			undefined,
+			recorder,
+		).fetchOrgBearerToken(globalToken, this.activeOrgId);
+		this.activeOrgToken = orgToken;
+		await this.setActiveOrg(this.activeOrgId, orgToken);
+	}
+
+	private async validateConnectionWithOrgRetry(
```
Rifdhan: There is no retry anymore, can call it validateConnection

src/servers/mcp-server.ts
```diff
@@ -35,14 +43,189 @@ import {
...
+			}
+			if (this.activeOrgId && !this.activeOrgToken) {
+				await this.forceRecreateActiveOrgToken();
+			}
+		} catch (error) {
+			console.error("Failed to load active org on connect:", error);
+		}
+	}
+
+	private async loadOrSeedWarmToken(): Promise<void> {
```
Rifdhan: Let's call it initGlobalTokenAndReconcileWithStorage

src/servers/mcp-server.ts
```diff
@@ -612,19 +829,102 @@ Provide this url to the user as a link to view the liveboard in ThoughtSpot.`;
 		);
 	}
 
+	@WithSpan("call-list-orgs")
+	async callListOrgs(recorder: MetricsRecorder) {
+		const span = trace.getSpan(context.active());
+
+		await this.loadOrSeedWarmToken();
```
Rifdhan: Why do we need to call loadOrSeedWarmToken here again? Isn't it always called in postInit when creating the McpServer class?

src/servers/mcp-server.ts
```diff
@@ -612,19 +829,102 @@ Provide this url to the user as a link to view the liveboard in ThoughtSpot.`;
...
+		recorder: MetricsRecorder,
+	) {
+		const span = trace.getSpan(context.active());
+		const { org_id } = SwitchOrgInputSchema.parse(request.params.arguments);
+		const orgId = String(org_id);
+		span?.setAttribute("requested_org_id", orgId);
+
+		let orgToken: string;
+		try {
+			await this.loadOrSeedWarmToken();
```
Rifdhan: Ditto

src/servers/mcp-server.ts
```diff
@@ -612,19 +829,102 @@ Provide this url to the user as a link to view the liveboard in ThoughtSpot.`;
...
+			const globalToken = this.globalToken ?? this.ctx.props.accessToken;
```
Rifdhan: loadOrSeedWarmToken already sets globalToken to accessToken as a fallback, why do we need to do it again here?

src/servers/user-token-store-server.ts
```diff
@@ -0,0 +1,280 @@
...
+						orgToken: orgToken ?? null,
+					});
+				}
+
+				// Clear the token only on a real org change; re-setting the same org
+				// (every cold connect) must not delete a token a sibling just minted.
+				case "POST /active-org-id-and-token": {
+					const body = (await request.json()) as {
+						activeOrgId: string;
+						orgToken?: string | null;
```
Rifdhan: Let's keep the naming scheme: activeOrgToken

src/servers/user-token-store-server.ts
```diff
@@ -0,0 +1,280 @@
...
+					};
+					if (body.orgToken) {
+						await this.state.storage.put({
+							[ACTIVE_ORG_KEY]: body.activeOrgId,
+							[ORG_TOKEN_KEY]: body.orgToken,
+						});
+						return Response.json({ ok: true });
+					}
+					const previousOrgId =
```
Rifdhan, line 61: When would we reach this flow that the caller passed org ID but no org token? Shouldn't we always expect both to be present?

src/servers/user-token-store-server.ts
```diff
@@ -0,0 +1,280 @@
...
+						await this.state.storage.put<string>(
+							ACTIVE_ORG_KEY,
+							body.activeOrgId,
+						);
+						await this.state.storage.delete(ORG_TOKEN_KEY);
+					}
+					return Response.json({ ok: true });
+				}
+
+				case "POST /active-org-token": {
```
Rifdhan: To further simplify, we can get rid of this API and just have the caller use POST /active-org-id-and-token

src/servers/user-token-store-server.ts
```diff
@@ -0,0 +1,280 @@
...
+const GLOBAL_TOKEN_REFRESH_INTERVAL_MS = 11 * 60 * 60 * 1000;
+const SESSION_IDLE_TTL_MS = 14 * 24 * 60 * 60 * 1000;
+// Org-token validity (24h); re-minted alongside the global token each refresh.
+const ORG_TOKEN_VALIDITY_SEC = 24 * 60 * 60;
+
+export type GlobalTokenData = {
+	globalToken: string;
+	globalRefreshToken: string;
+	instanceUrl: string;
+	expiresAt?: number;
```
Rifdhan: globalTokenExpiresAt

src/servers/user-token-store-server.ts
```diff
@@ -0,0 +1,280 @@
...
+			return Response.json({ error: message }, { status: 500 });
+		}
+	}
+
+	private async seedGlobalToken(store: GlobalTokenData): Promise<void> {
+		const existing =
+			await this.state.storage.get<GlobalTokenData>(GLOBAL_TOKEN_KEY);
+		const existingAlarm = await this.state.storage.getAlarm();
+		if (existing?.globalToken === store.globalToken) {
+			if (existingAlarm == null) {
```
Rifdhan: We always set an alarm when we set a global token, so this scenario would never be possible. I want to clean up unnecessary or impossible code paths to simplify.

src/servers/user-token-store-server.ts
```diff
@@ -0,0 +1,280 @@
...
+	}
+
+	private async touchLastSeen(): Promise<void> {
+		const store =
+			await this.state.storage.get<GlobalTokenData>(GLOBAL_TOKEN_KEY);
+		if (!store) {
+			return;
+		}
+		const now = Date.now();
+		const THROTTLE_MS = 60 * 60 * 1000; // 1 hour
```
Rifdhan, line 128: Do we really need a throttle? Let's simplify

src/servers/user-token-store-server.ts
```diff
@@ -0,0 +1,280 @@
...
+		if (store.lastSeenAt && now - store.lastSeenAt < THROTTLE_MS) {
+			return;
+		}
+		await this.state.storage.put<GlobalTokenData>(GLOBAL_TOKEN_KEY, {
+			...store,
+			lastSeenAt: now,
+		});
+	}
+
+	private async refreshGlobalToken(): Promise<void> {
```
Rifdhan: refreshGlobalAndOrgTokens

src/storage-service/storage-service.ts
```diff
@@ -1,21 +1,10 @@
...
- *
- * Communicates directly with the DO via its stub (bypassing the OAuth layer), mapping to the
- * following HTTP endpoints exposed by the server:
- *   POST  /storage/<storageId>/initialize —> initializeConversation
- *   POST  /storage/<storageId>/append     —> appendMessagesAndRestartTtl
- *   GET   /storage/<storageId>/messages   —> getNewMessagesAndUpdateBookmark
- *
- * The storageId is derived by taking a hash of the user's access token and combining it with the
- * conversationId, to ensure no users can access each other's conversations.
- */
```
Rifdhan, line 14: Why did we remove this comment?

src/storage-service/storage-service.ts
```diff
@@ -32,16 +21,148 @@ export class StorageServiceClient {
...
 	}
 
-	/**
-	 * Initialize a conversation. Must be called before appending messages.
-	 * Can also be called on an existing conversation that is already marked done,
-	 * to prime it for a follow-up message.
-	 */
+	private static readonly ACTIVE_ORG_ID = "__active_org__";
+
+	async getActiveOrg(): Promise<{
```
Rifdhan: Let's put all the org-related utils in a separate service file as discussed

src/utils.ts
```diff
@@ -3,11 +3,16 @@ import {
...
 	McpServerError as PkgMcpServerError,
 } from "@thoughtspot/mcp-auth";
-import type { ApiVersionMode } from "./metrics/runtime/metric-types";
+import type { ApiVersionMode, AuthMode } from "./metrics/runtime/metric-types";
 import { getActiveSpan } from "./metrics/tracing/tracing-utils";
 
 export type Props = {
 	accessToken: string;
+	// From gettoken, for keep-warm refresh.
+	refreshToken?: string;
```
Rifdhan: globalRefreshToken

src/utils.ts
```diff
@@ -3,11 +3,16 @@ import {
...
+	tokenCreatedTime?: number;
```
Rifdhan: globalTokenCreatedAt

src/utils.ts
```diff
@@ -3,11 +3,16 @@ import {
...
+	// Absolute epoch-ms expiry (despite the name).
+	tokenExpiryDuration?: number;
```
Rifdhan: globalTokenExpiresAt (let's fix the name as per comment)

static/oauth-callback.js
Rifdhan: We should error in this path - the user now MUST paste the full JSON and not just the token string. Similarly let's update the UI text in oauth-callback.html to tell the user to copy the entire JSON

Image

src/servers/mcp-server.ts
```diff
@@ -35,14 +43,189 @@ import {
...
 
+	protected getActiveOrgId(): string | undefined {
+		return this.activeOrgId;
+	}
+
+	protected getActiveToken(): string {
+		if (this.activeOrgId) {
+			if (!this.activeOrgToken) {
+				throw new Error(
+					`Org ${this.activeOrgId} is active but its token is not minted`,
```
mouryabalabhadra, line 64: **[correctness, high] getActiveToken() throws → poisons the whole session, not just org tools.**

OAuth connect on an orgs-enabled cluster: `postInit()` sets `activeOrgId` via `setActiveOrg(currentOrgId)` (stores the id with **no token**) then calls `forceRecreateActiveOrgToken()`. A transient 5xx on that mint is swallowed by postInit's try/catch, leaving `activeOrgId` set + `activeOrgToken` undefined in storage. On the next `callTool`, `loadActiveOrg()` rehydrates `{orgId, token:null}` → `getThoughtSpotService()` → `getActiveToken()` throws here. `get_answer` / `create_analysis_session` / `create_dashboard` all raw-throw; `check_connectivity` throws too (it hits `getActiveToken` before validating). No reactive re-mint path — the session stays broken until the 11h keep-warm alarm re-mints.

Consider minting before persisting the org id, or falling back to the global token (with a re-mint) instead of throwing.

src/servers/mcp-server.ts
```diff
@@ -189,10 +379,18 @@ export class MCPServer extends BaseMCPServer {
...
+
+		if (this.areOrgToolsAvailable()) {
+			await this.loadActiveOrg();
+		}
+
 		switch (name) {
 			case ToolName.Ping: {
 				if (this.ctx.props.accessToken && this.ctx.props.instanceUrl) {
-						if (!this.getThoughtSpotService(recorder).validateConnection()) {
+					if (!this.getThoughtSpotService(recorder).validateConnection()) {
```
mouryabalabhadra: **[correctness, high] `validateConnection()` is not awaited → Ping always reports success.**

`validateConnection()` is `async` (returns `Promise<boolean>`). `!<pending Promise>` is always `false`, so the failure branch never runs and Ping returns `Pong` even against an expired token / unreachable instance. This PR fixed the identical sibling call in `check_connectivity` (now `await this.validateConnectionWithOrgRetry(...)`) but left this one un-awaited, so the two connectivity tools diverge — Ping is a false-positive health signal.

```suggestion
					if (!(await this.getThoughtSpotService(recorder).validateConnection())) {
```

src/servers/mcp-server.ts
```diff
@@ -35,14 +43,189 @@ import {
...
+
+	private async validateConnectionWithOrgRetry(
+		recorder?: MetricsRecorder,
+	): Promise<boolean> {
+		const usedOrgToken = Boolean(this.activeOrgId && this.activeOrgToken);
+		const ok = await this.getThoughtSpotService(recorder).validateConnection();
+		if (ok || !usedOrgToken) {
+			return ok;
+		}
+		await this.forceRecreateActiveOrgToken(recorder);
```
mouryabalabhadra: **[correctness] Unguarded `forceRecreateActiveOrgToken()` → `check_connectivity` throws a raw error instead of the graceful response.**

When the user was switched into an org they later lost access to, the first `validateConnection()` returns false, so this re-mint runs and `fetchOrgBearerToken` throws `ThoughtSpotApiError(403)`. There is no try/catch here or in `callTool`, so it propagates to the SDK handler and surfaces as a raw JSON-RPC error rather than the intended `createErrorResponse('Failed to validate connection', ...)`. Wrap the re-mint and fall through to `return false` on failure.

src/servers/mcp-server-base.ts
```diff
@@ -187,23 +202,53 @@ export abstract class BaseMCPServer extends Server {
...
+	 * conversation buffers and active-org state), keeping users isolated.
+	 *
+	 * Keyed on the refresh token when present (OAuth): it is stable across the
+	 * access token's 24h rotation and only changes on full reauthentication, so
+	 * storage survives token refresh and resets on reauth. Falls back to the
+	 * access token for static bearer/token connections, which have no refresh
+	 * token (their token is long-lived).
+	 */
+	protected async getStorageKeyHash(): Promise<string> {
+		const keyToken = this.ctx.props.refreshToken ?? this.ctx.props.accessToken;
```
mouryabalabhadra: **[durability, plausible] Keying durable storage on `refreshToken` assumes refresh tokens never rotate.**

The comment asserts the refresh token 'only changes on full reauthentication', but `extendGrantProps` stamps `refreshToken` from each token exchange and `refreshGlobalToken` explicitly reads a rotated `data.refreshToken` back — evidence they do rotate. If a new refresh token is issued on periodic re-grant, this hash changes, so `switch_org`'s persisted selection (and buffered conversations) becomes unreachable and resets to defaults — contradicting the `switch_org` tool description's 'persists across reconnects / shared across sessions' promise. Also orphans conversation buffers written under the old accessToken-hash scheme at the deploy boundary. Worth confirming ThoughtSpot's refresh-token rotation policy.

src/servers/user-token-store-server.ts
```diff
@@ -0,0 +1,280 @@
...
+		}
+	}
+
+	// Re-mint the active org token from a freshly refreshed global token, so it
+	// stays as warm as the global token (both 24h). No-op if no org is active.
+	// On failure, keep the existing org token and let the next 11h alarm retry;
+	// never throws (the global refresh must still commit and the alarm re-arm).
+	// A token that is genuinely expired only surfaces after 14 idle days, as an
+	// expected 401 — there is deliberately no reactive re-mint / retry path.
+	private async refreshOrgToken(
```
mouryabalabhadra, line 197: **[reuse/altitude] DO hand-rolls org-token + gettoken fetch already implemented in `thoughtspot-client.ts`.**

`refreshOrgToken()` / `refreshGlobalToken()` duplicate the endpoint path, headers, `data?.data?.token ?? data?.token` extraction, and the `ORG_TOKEN_VALIDITY_SEC` constant that `fetchOrgBearerToken` / `buildHeaders` already own. The fragile private-endpoint knowledge ('/callosum/v2/... 404s; token nested under data.token') now lives in two files. A fix to the client won't propagate here, so interactive `switch_org` keeps working while every background 11h keep-warm re-mint silently mints against a stale path — org tokens go stale under long-lived sessions with no signal until a 401 days later. Route the DO refresh through the shared client / `OrgService`.

mouryabalabhadra (commented):
Automated code review (medium effort, 8 finder angles). 3 confirmed correctness bugs, 1 plausible durability risk, plus altitude/reuse/dead-code/efficiency cleanups. Inline comments below, ranked by severity.

src/servers/mcp-server.ts
```diff
@@ -35,14 +43,189 @@ import {
...
+			recorder,
+		).fetchOrgBearerToken(globalToken, this.activeOrgId);
+		this.activeOrgToken = orgToken;
+		await this.setActiveOrg(this.activeOrgId, orgToken);
+	}
+
+	private async validateConnectionWithOrgRetry(
+		recorder?: MetricsRecorder,
+	): Promise<boolean> {
+		const usedOrgToken = Boolean(this.activeOrgId && this.activeOrgToken);
```
mouryabalabhadra: **[altitude] Stale-org-token recovery is special-cased to `check_connectivity` only.**

`validateConnectionWithOrgRetry` (re-mint then re-validate) is bolted onto one diagnostic tool; all data tools call `getActiveToken()` with no recovery. If the org token expires between keep-warm re-mints (the 11h alarm hit its 'will retry next interval' failure path, or the 14-idle-day boundary), `create_analysis_session` / `send_session_message` / `create_dashboard` / `getDatasources` all 401 the user, while `check_connectivity` on the same session transparently recovers. Push the re-mint-on-401 to the client/service layer (or `getActiveToken`) so every tool benefits.

src/storage-service/storage-service.ts
```diff
@@ -32,16 +21,148 @@ export class StorageServiceClient {
...
+				body: JSON.stringify({ activeOrgId, orgToken: orgToken ?? null }),
+			},
+		);
+		if (!response.ok) {
+			const body = await response.text();
+			throw new Error(`Failed to set active org (${response.status}): ${body}`);
+		}
+	}
+
+	async setActiveOrgToken(orgToken: string): Promise<void> {
```
mouryabalabhadra: **[cleanup] Dead code: `setActiveOrgToken()` and its DO route `POST /active-org-token` have no production caller.**

The org token is only ever written via `setActiveOrg`'s combined `active-org-id-and-token` path; this method and the corresponding case in `user-token-store-server.ts` are referenced only by tests. Delete both, or wire them in.

src/servers/mcp-server.ts
```diff
@@ -189,10 +379,18 @@ export class MCPServer extends BaseMCPServer {
 		const { name } = request.params;
 		this.trackers.track(TrackEvent.CallTool, { toolName: name });
 
+		if (this.ctx.props.authMode === "oauth") {
+			this.touchLastSeen();
+		}
+
+		if (this.areOrgToolsAvailable()) {
```
mouryabalabhadra, line 435: **[efficiency] `loadActiveOrg()` does a blocking DO round-trip on every `callTool`.**

Active-org state only changes on `switch_org` and is already held in instance fields, yet each tool call (including `get_session_updates` poll loops) computes a SHA-256 and awaits a UserTokenStore DO fetch — adding the DO's latency/availability as a new precondition to every invocation. `list_orgs` reloads it 2–3× (here, `loadOrSeedWarmToken`, then again in `callListOrgs`). Load once in `postInit`, mutate only in `setActiveOrg`, and drop the redundant reloads (or gate behind a short in-memory TTL to still observe sibling-session switches).

src/servers/mcp-server-base.ts
rohitthughtspot: @mouryabalabhadra no refresh token is static and does not rotate okta does provide a way to do it but currently we are using static refresh tokens without refresh and i don't think that will change

src/servers/mcp-server.ts
rohitthughtspot: added a check in tool call to verify whether token exist if not then remint token

src/servers/mcp-server.ts
rohitthughtspot: The active org is shared across the user's sessions via the token DO, so reload per call — a switch_org in one session must be seen by the others on their next tool call (an instance-cached value would go stale).

added a comment in file

src/servers/mcp-server.ts
rohitthughtspot: fixed

src/servers/user-token-store-server.ts
rohitthughtspot: done

src/servers/user-token-store-server.ts
rohitthughtspot: This is basically a switch org race condition when 2 concurrent switch orgs happen we need to be consistent

src/servers/user-token-store-server.ts
rohitthughtspot: done

src/servers/user-token-store-server.ts
rohitthughtspot: removed

src/servers/user-token-store-server.ts
rohitthughtspot: Called on every tool call, but lastSeenAt only feeds the 14-day idle TTL, so throttle to ~1 write/hour to avoid a DO write on every call.

src/servers/user-token-store-server.ts
rohitthughtspot: done

src/utils.ts
rohitthughtspot: done

src/utils.ts
rohitthughtspot: done

src/utils.ts
rohitthughtspot: done

static/oauth-callback.js
rohitthughtspot: it is dependent on ts-mcp-auth can't be changed without changing ts-mcp-auth

static/oauth-callback.js
rohitthughtspot: done

src/servers/user-token-store-server.ts
rohitthughtspot: done removed it

src/servers/mcp-server.ts
rohitthughtspot: done

src/servers/mcp-server.ts
rohitthughtspot: done

src/servers/mcp-server.ts
rohitthughtspot: removed from here as well missed this tool

src/servers/mcp-server.ts
rohitthughtspot: fixed it

src/servers/user-token-store-server.ts
```diff
@@ -0,0 +1,223 @@
...
+					return new Response("Not Found", { status: 404 });
+			}
+		} catch (err) {
+			const message = err instanceof Error ? err.message : String(err);
+			console.error("Error handling user token-store request:", message);
+			return Response.json({ error: message }, { status: 500 });
+		}
+	}
+
+	private async seedGlobalToken(store: GlobalTokenData): Promise<void> {
```
Rifdhan: storeGlobalTokenData

src/servers/user-token-store-server.ts
```diff
@@ -0,0 +1,223 @@
...
+			return;
+		}
+
+		try {
+			const refreshed = await refreshGlobalToken({
+				instanceUrl: store.instanceUrl,
+				globalToken: store.globalToken,
+				globalRefreshToken: store.globalRefreshToken,
+			});
+			await this.state.storage.put<GlobalTokenData>(GLOBAL_TOKEN_KEY, {
```
Rifdhan, line 164: Can we call seedGlobalToken to avoid repeating some of the below logic

src/storage-service/org-storage-service.ts
```diff
@@ -0,0 +1,134 @@
+// Client for the per-user org/token Durable Object (UserTokenStoreSQLite,
+// bound as USER_TOKEN_OBJECT). Owns the active-org selection and the keep-warm
+// global cluster token. Kept separate from the Spotter conversation storage
+// (StorageServiceClient) so the two surface areas stay isolated.
+export class OrgStorageServiceClient {
+	// Single per-user DO instance name; all org/token state lives under it.
+	private static readonly ACTIVE_ORG_ID = "__active_org__";
```
Rifdhan: This variable name can be confused with the ID of the active org itself. Let's call this STORAGE_STUB_ID.

src/storage-service/org-storage-service.ts
```diff
@@ -0,0 +1,134 @@
...
+			`${this.accessTokenHashUrlSafe}:${id}`,
+		);
+		return this.namespace.get(doId);
+	}
+
+	private url(id: string, operation: string): string {
+		return `https://internal/storage/${encodeURIComponent(id)}/${operation}`;
+	}
+
+	async getActiveOrg(): Promise<{
```
Rifdhan: Let's name the utils with the same scheme as the request paths. So for this one getActiveOrgIdAndToken, etc

src/storage-service/storage-service.ts
Rifdhan: There are no code changes here, let's revert the changes in this file

src/servers/mcp-server.ts
rohitthughtspot: because list orgs needs to read token from do and postintit might not have most recent token

src/servers/mcp-server.ts
rohitthughtspot: same

src/servers/user-token-store-server.ts
rohitthughtspot: done

src/servers/user-token-store-server.ts
rohitthughtspot: hey org tokens are refreshed every 11 hours and are loaded from do store on tool calls same for global tokens

src/storage-service/org-storage-service.ts
rohitthughtspot: done

src/storage-service/org-storage-service.ts
rohitthughtspot: done

src/servers/user-token-store-server.ts
rohitthughtspot: we can't reuse it here. storeGlobalTokenData skips its work if the token hasn't changed, but gettoken often returns the same token, so that would skip resetting the alarm and the keep-warm refresh would just stop, we can do this later as a seperate pr

Rifdhan (approved):

mouryabalabhadra (approved):

PR #177, opened 2026-07-08
[SCAL-319676] changes for `search_objects` and `fetch_data` tools, along with local instance deployment

saharsh-ts:


patches/@thoughtspot+mcp-auth+1.0.0.patch
mouryabalabhadra, line 1: patches/@thoughtspot+mcp-auth+1.0.0.patch should not be there

PR #178, opened 2026-07-09
[SCAL-323294] changes for search_objects and added terminologies

saharsh-ts:


mouryabalabhadra:
@copilot please review

src/servers/tool-definitions.ts
```diff
@@ -65,6 +65,123 @@ export const GetAnswerOutputSchema = z.object({
 		.describe("Information about the fields in the answer"),
 });
 
+export const SearchObjectsInputSchema = z.object({
+	query: z
+		.union([z.string().min(1), z.array(z.string().min(1)).nonempty()])
+		.describe(
+			"The search term(s) to find objects for, matched against object names or descriptions. Pass a single string for one search, or an array of distinct terms when the user's request covers several separate things. Each term is searched in parallel and the results are merged and de-duplicated. Note: pagination via `cursor` only applies to a single-term search.",
```
Rifdhan: Is there a prominent need to support multiple disjoint search terms? I feel a single search term will be the 95% case and would prefer simplifying the contract to keep it simple for agent understanding. Especially since some features (eg cursor) won't work in that mode, it can lead to more problems.

src/servers/tool-definitions.ts
```diff
@@ -65,6 +65,123 @@ export const GetAnswerOutputSchema = z.object({
...
+		),
+	types: z
+		.array(z.string())
+		.optional()
+		.describe(
+			"Restrict results to these object types. Accepts friendly and legacy names: 'liveboard' (aliases 'pinboard', 'dashboard'), 'answer', and 'worksheet' (aliases 'logical table', 'data model', 'data source'). Omit to search all types.",
+		),
+	owner: z
```
Rifdhan: Should we call the field author_name instead of owner

src/servers/tool-definitions.ts
```diff
@@ -65,6 +65,123 @@ export const GetAnswerOutputSchema = z.object({
...
+		.string()
+		.optional()
+		.describe(
+			"Restrict results to objects authored by this user, matched against the author's display name (case-insensitive).",
```
Rifdhan: Nit: prefer "authored by the provided user" rather than "authored by this user" to avoid confusing with the end user themself

src/servers/tool-definitions.ts
```diff
@@ -65,6 +65,123 @@ export const GetAnswerOutputSchema = z.object({
...
+	cursor: z
+		.string()
+		.optional()
+		.describe(
+			"Opaque pagination cursor returned as `next_cursor` by a previous call. Omit for the first page.",
+		),
+});
+
+const SearchObjectHeaderSchema = z.object({
+	id: z
```
Rifdhan: Let's call it object_id for consistency

src/servers/tool-definitions.ts
```diff
@@ -65,6 +65,123 @@ export const GetAnswerOutputSchema = z.object({
...
+		.describe(
+			"The GUID to pass to fetch_data as `object_id`. For a visualization pinned on a Liveboard this is the parent Liveboard's GUID (see `visualization_id`).",
+		),
+	visualization_id: z
+		.string()
+		.optional()
+		.describe(
+			"Set only when this result is a specific visualization on a Liveboard: `id` is the Liveboard and this is the visualization. Pass it to fetch_data as `visualization_ids` to fetch just this viz.",
+		),
+	name: z.string().describe("The display name/title of the object."),
```
Rifdhan: Let's call it title for consistency with other schemas

src/servers/tool-definitions.ts
```diff
@@ -65,6 +65,123 @@ export const GetAnswerOutputSchema = z.object({
...
+		.number()
+		.nullable()
+		.optional()
+		.describe(
+			"Epoch-millisecond timestamp the current user last viewed the object (viewedbymeTime). Null when the search backend does not expose it.",
+		),
+	verified: z.boolean().describe("Whether the object is marked as verified."),
+	frame_url: z
+		.string()
+		.describe("Deep link to open the object in the ThoughtSpot UI."),
```
Rifdhan: Is this embeddable or meant to be opened in a new tab?

src/servers/tool-definitions.ts
```diff
@@ -65,6 +65,123 @@ export const GetAnswerOutputSchema = z.object({
...
+	next_cursor: z
+		.string()
+		.nullable()
+		.describe(
+			"Cursor to pass back as `cursor` to fetch the next page, or null when there are no more results.",
+		),
+	request_id: z
+		.string()
+		.describe(
+			"Correlation id sent on the upstream call as x-request-id; trace this in ThoughtSpot's server logs.",
```
Rifdhan: Why is this being sent back? It seems like noise that is not useful to the host agent

src/thoughtspot/search-objects/search-objects-query.ts
```diff
@@ -0,0 +1,82 @@
+// GraphQL query for the private Eureka object search (no public REST equivalent),
```
Rifdhan, line 1: Let's follow the kebab-case naming convention for files and folders. I also don't think we need a grpc nested folder. We can put this code under `src/thoughtspot/search-objects/*`

src/thoughtspot/search-objects/search-objects.ts
```diff
@@ -0,0 +1,394 @@
...
+				{
+					operationName: "GetEurekaResults",
+					query: searchObjectsQuery,
+					variables: {
+						params: {
+							batchSize: limit,
+							// STICKERS facet lets tag ids resolve to names.
+							desiredFacets: [{ facetType: "STICKERS", facetValue: [] }],
+							facetSelections,
+							maxPinboardVizCount: 5,
```
Rifdhan: How did we decide on this value? An agent may think there are no more results if they requested limit=10 and only got 5 results (if there were only pinboard viz results for the search).

src/thoughtspot/grpc/grpc-utils.ts
```diff
@@ -0,0 +1,53 @@
+// Shared helpers for the gRPC-backed ThoughtSpot tools.
+
+// Fresh id sent as x-request-id so a call can be traced across systems.
+export function generateRequestId(): string {
+	return globalThis.crypto.randomUUID();
+}
+
+// Standard headers for ThoughtSpot HTTP calls. `acceptLanguage` is opt-in: the
+// gRPC/Eureka endpoints 500 without it, but the legacy calls never sent it.
+export function buildTsHeaders(
```
Rifdhan: This is duplicate, use `buildHeaders` existing util and extend if needed

src/thoughtspot/terminology.ts
```diff
@@ -0,0 +1,87 @@
...
+	canonical: string;
+	// Informal names and legacy names users may use for the same concept.
+	synonyms: string[];
+}
+
+// Object glossary; the synonym lookups below are derived from it.
+export const TERMINOLOGY: TerminologyEntry[] = [
+	{
+		canonical: "Liveboard",
+		synonyms: ["pinboard", "dashboard", "board"],
```
Rifdhan: Why is canonical capitalized but synonyms are not? Should we make it consistent

src/thoughtspot/thoughtspot-service.ts
```diff
@@ -743,17 +747,40 @@ export class ThoughtSpotService {
...
 	@WithSpan("validate-connection")
 	async validateConnection(): Promise<boolean> {
 		try {
-			await this.observeUpstreamCall(
+			const info = await this.observeUpstreamCall(
 				UPSTREAM_OPERATION_NAMES.getSessionInfo,
 				() => (this.client as any).getSessionInfo(),
 			);
-			return true;
+			return Boolean(info?.userGUID);
```
Rifdhan, line 783: What is the reason for this change?

src/servers/tool-definitions.ts
saharsh-ts: supporting only single search term/string (can be multi-word, e.g. "sales by region")

src/servers/tool-definitions.ts
saharsh-ts: done

src/servers/tool-definitions.ts
saharsh-ts: done

src/servers/tool-definitions.ts
saharsh-ts: done

src/servers/tool-definitions.ts
saharsh-ts: done

src/servers/tool-definitions.ts
saharsh-ts: open in new tab

src/thoughtspot/search-objects/search-objects-query.ts
saharsh-ts: done

src/thoughtspot/grpc/grpc-utils.ts
saharsh-ts: done

src/servers/tool-definitions.ts
saharsh-ts: for debugging purpose only. can be removed if you'd like

src/thoughtspot/thoughtspot-service.ts
saharsh-ts: `validateConnection` used to return true whenever `getSessionInfo` didn't throw — but that call can return 200 with an empty/invalid body (e.g. an unauthenticated session), reporting healthy when it isn't. 

It now returns Boolean(info?.userGUID) to confirm a real authenticated session.

src/thoughtspot/terminology.ts
saharsh-ts: updated in `object-types.ts`

src/thoughtspot/search-objects/search-objects.ts
saharsh-ts: it got picked as the default values from the eureka apis.

On the "agent thinks there's nothing more" concern: `next_cursor` is driven by the backend's `isFinalPage/totalResults`, not by how many rows come back. So a short page never on its own signals "no more results."

src/servers/mcp-server.ts
```diff
@@ -955,6 +932,54 @@ Provide this url to the user as a link to view the liveboard in ThoughtSpot.`;
...
+				owner: author_name,
+				tag,
+				modifiedSince: modified_since,
+				verifiedOnly: verified_only,
+				limit,
+				cursor,
+			});
+
+			// no_results and error are returned as structured content (not a
+			// protocol error) so the model gets the typed envelope to render.
```
Rifdhan: What does "typed envelope to render" mean? This comment is a bit unclear to me.

src/servers/tool-definitions.ts
```diff
@@ -65,6 +65,163 @@ export const GetAnswerOutputSchema = z.object({
...
+		.describe(
+			"UPPER-case object type: 'LIVEBOARD', 'ANSWER', 'LIVEBOARD_VIZ' (a viz pinned on a Liveboard, i.e. a 'Liveboard viz'; `object_id` is the parent Liveboard and `visualization_id` the viz) or 'WORKSHEET'. Any of these can be passed straight back as a `types` filter.",
+		),
+	author_name: z
+		.string()
+		.describe("The display name of the user who authored the object."),
+	description: z
+		.string()
+		.nullable()
+		.describe("The description of the object, or null when it has none."),
```
Rifdhan: Prefer to make this optional and omit if not present, rather than nullable. This is how we handle other fields in other places.

src/servers/tool-definitions.ts
```diff
@@ -65,6 +65,163 @@ export const GetAnswerOutputSchema = z.object({
...
+	tags: z
+		.array(z.string())
+		.describe("The names of the tags/stickers applied to the object."),
+	last_modified: z
+		.string()
+		.nullable()
+		.describe(
+			"ISO-8601 timestamp of the last modification (e.g. 2026-05-15T14:30:00Z), or null when unavailable. Render as a plain date.",
```
Rifdhan: Ditto

src/servers/tool-definitions.ts
Rifdhan: Got it, prefer to call this field something like external_link

src/servers/tool-definitions.ts
```diff
@@ -65,6 +65,163 @@ export const GetAnswerOutputSchema = z.object({
...
+		.nullable()
+		.optional()
+		.describe(
+			"Cursor to pass back as `cursor` for the next page; null when there are no more results, and omitted on error.",
+		),
+	message: z
+		.string()
+		.optional()
+		.describe(
+			"Present on no_results: human-readable 'nothing matched' note; safe to relay to the user.",
```
Rifdhan: Why do we need this? The host agent can display a reasonable no results message themselves.

src/servers/tool-definitions.ts
Rifdhan: Let's remove it, we want to minimize the tool schema surface area as much as possible

src/thoughtspot/search-objects/search-objects.ts
```diff
@@ -0,0 +1,481 @@
...
+	SearchErrorCode,
+	SearchObjectHeader,
+	SearchObjectResult,
+	SearchObjectsError,
+	SearchObjectsParams,
+	SearchObjectsResponse,
+} from "./search-objects-types";
+
+// Build a deep link to the object in the ThoughtSpot UI from its result type.
+function buildFrameUrl(
```
Rifdhan: We have inconsistent naming for utils in this file, eg "buildFrameUrl", "toEpochMs", "deriveQuery". Let's use a conistent verb if possible.

src/thoughtspot/search-objects/search-objects.ts
```diff
@@ -0,0 +1,481 @@
...
+			facetSelections.push({
+				facetType: "OBJECT_TYPE",
+				facetValue: resolveObjectTypeFacets(types),
+			});
+		}
+		if (verifiedOnly) {
+			facetSelections.push({ facetType: "IS_VERIFIED", facetValue: ["true"] });
+		}
+
+		const endpoint = "/prism/?op=GetEurekaResults";
```
Rifdhan: Inline this

src/thoughtspot/search-objects/search-objects.ts
```diff
@@ -0,0 +1,481 @@
...
+						confidence: result?.score ?? 0,
+					};
+				})
+				.filter(
+					(obj: SearchObjectHeader | null): obj is SearchObjectHeader =>
+						obj !== null,
+				);
+
+			if (owner) {
+				const needle = owner.toLowerCase();
```
Rifdhan: What does needle mean?

src/servers/mcp-server.ts
saharsh-ts: simplified it. it was basically to say that no_results/error are returned as normal structured content (not an MCP protocol error) so the client still gets the structured response.

src/servers/tool-definitions.ts
saharsh-ts: description is optional now, omitted when absent.

src/servers/tool-definitions.ts
saharsh-ts: same, last_modified optional/omitted.

src/thoughtspot/search-objects/search-objects.ts
saharsh-ts: consistent to* naming, moved to a mapper module.

src/servers/tool-definitions.ts
saharsh-ts: dropped the no-results message; the host agent renders its own.

src/thoughtspot/search-objects/search-objects.ts
saharsh-ts: done

src/thoughtspot/search-objects/search-objects.ts
saharsh-ts: Renamed - it's the lower-cased substring for the post-filter match; now ownerLower/tagLower.

Rifdhan (approved):

PR #179, opened 2026-07-09
[SCAL-323294] changes for fetch_data tool

saharsh-ts:


mouryabalabhadra:
@copilot review the pr

saharsh-ts:
@copilot review the pr

src/servers/tool-definitions.ts
```diff
@@ -65,6 +65,241 @@ export const GetAnswerOutputSchema = z.object({
...
+		),
+	viz_name: z
+		.string()
+		.optional()
+		.describe(
+			"The name of the visualization. Present for Liveboard visualizations.",
+		),
+	columns: z.array(z.string()).describe("The column names, in order."),
+	data_rows: z
+		.array(z.array(z.unknown()))
```
Rifdhan: Is there a way to add a proper type contract

src/thoughtspot/fetch-data/fetch-data.ts
```diff
@@ -0,0 +1,207 @@
...
+	visualization_name?: string;
+}
+
+function isObjectRow(row: unknown): row is Record<string, unknown> {
+	return typeof row === "object" && row !== null && !Array.isArray(row);
+}
+
+// Round numeric cells to 2 decimals: collapses FP noise and trims payload.
+// Non-numbers and non-finite values pass through untouched.
+function roundCell(value: unknown): unknown {
```
Rifdhan: This seems arbitrary, what if the customer wants more precision? I feel we should return the full value for now, and not prematurely optimize this.

src/thoughtspot/fetch-data/fetch-data.ts
```diff
@@ -0,0 +1,207 @@
...
+			throw new Error(
+				`fetchData does not support object type "${objectType}" (id ${objectId}); only Answers and Liveboards expose fetchable data.`,
+			);
+		}
+
+		// `record_size` caps returned rows for Answers, but the Liveboard endpoint
+		// treats it as "must hold the whole viz" and 500s when it's too small.
+		// Start at maxRows; on that error, bump to the required count and refetch,
+		// then cap rows client-side below. Attempts are bounded (a full Liveboard
+		// can report a larger viz on each retry) to avoid a runaway loop.
```
Rifdhan: This seems like a hacky design, we should make the API call such that it is successful on the first try. Is there no way to do that using this API?

src/thoughtspot/fetch-data/fetch-data.ts
saharsh-ts: Removed — cells now returned at full precision.

src/servers/tool-definitions.ts
saharsh-ts: Tightened to string | number | boolean | null datatypes.

src/servers/mcp-server.ts
```diff
@@ -192,7 +194,7 @@ export class MCPServer extends BaseMCPServer {
 		switch (name) {
 			case ToolName.Ping: {
 				if (this.ctx.props.accessToken && this.ctx.props.instanceUrl) {
-						if (!this.getThoughtSpotService(recorder).validateConnection()) {
+					if (!this.getThoughtSpotService(recorder).validateConnection()) {
```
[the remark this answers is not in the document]
saharsh-ts: added await

src/thoughtspot/fetch-data/fetch-data.ts
saharsh-ts: Reworked to one call — Answers pass record_size = max_rows, 
Liveboards request the max 32-bit int.
Verified on local instance.

saharsh-ts:
@copilot review the PR

src/servers/tool-definitions.ts
```diff
@@ -550,6 +629,22 @@ export const toolDefinitionsV2 = [
 			openWorldHint: false,
 		},
 	},
+	{
+		name: ToolName.FetchData,
+		description: [
+			"Fetch the full data (columns and rows) of a saved Answer or Liveboard, identified by its GUID.",
+			"CALL THIS WHENEVER the user wants to explain, describe, summarize, analyze, interpret, or ask what a specific object contains or shows: any explanation of what an object contains must be grounded in the data this tool returns, so fetch it first with the object's `id`, then answer from the returned data.",
+			"Returns the object's data exactly as saved — it does not run new queries or change the object's question, filters, or columns. The result is shaped to the object's type: an Answer returns a single tabular result, a Liveboard returns one tabular result per visualization (each with its visualization id and name). To pull a single visualization pinned on a Liveboard, pass the Liveboard GUID as `object_id` and the visualization GUID in `visualization_ids`. Each result includes the column names and data rows. Use the optional `max_rows` to bound the rows returned per visualization (defaults to 25).",
```
Rifdhan: This description is a bit misleading, since TS will execute the saved query against the latest state of the data, so the tool may return different data compared to when the user first saved the answer. It WILL also run a new database query, so not sure why this says it does not run new queries.

src/servers/tool-definitions.ts
```diff
@@ -213,6 +213,84 @@ export const SearchObjectsResponseSchema = z.object({
...
+		),
+});
+
+export const FetchDataOutputSchema = z.object({
+	id: z.string().describe("The GUID of the fetched object."),
+	name: z.string().describe("The display name of the object."),
+	type: z
+		.string()
+		.describe("The resolved object type: either 'ANSWER' or 'LIVEBOARD'."),
+	description: z.string().describe("The description of the object."),
```
Rifdhan: There are a lot of fields here duplicated from the search_objects response. Let's make it more concise and only include truly NEW data which the agent doesn't already know.

src/servers/tool-definitions.ts
```diff
@@ -213,6 +213,84 @@ export const SearchObjectsResponseSchema = z.object({
...
+	data: z
+		.array(FetchDataVizSchema)
+		.describe(
+			"The object's data. A single entry for an Answer; one entry per visualization for a Liveboard.",
+		),
+	request_id: z
```
Rifdhan: Let's remove like last time

src/servers/tool-definitions.ts
```diff
@@ -213,6 +213,84 @@ export const SearchObjectsResponseSchema = z.object({
...
+	total_row_count: z
+		.number()
+		.optional()
+		.describe(
+			"Total rows available upstream, which may exceed the rows returned when capped by `max_rows`.",
+		),
+	row_count: z
+		.number()
+		.optional()
+		.describe("Number of rows actually returned in `data_rows`."),
```
Rifdhan: Isn't this redundant? The agent can see how many rows there are

src/thoughtspot/fetch-data/fetch-data.ts
```diff
@@ -0,0 +1,162 @@
...
+// Keep in sync with the `max_rows` description in tool-definitions.ts.
+export const FETCH_DATA_DEFAULT_MAX_ROWS = 25;
+
+// "Unbounded" record_size for Liveboards (they 500 if it can't hold the whole
+// viz). Max 32-bit signed int — the endpoint reads it as a GraphQL Int.
+export const LIVEBOARD_RECORD_SIZE = 2_147_483_647;
+
+// Only Answers and Liveboards expose fetchable data.
+const ANSWER_TYPE = "ANSWER";
+const LIVEBOARD_TYPE = "LIVEBOARD";
```
Rifdhan: Should these be an enum

src/thoughtspot/fetch-data/fetch-data.ts
```diff
@@ -0,0 +1,162 @@
...
+		objectId,
+		vizIds,
+		maxRows = FETCH_DATA_DEFAULT_MAX_ROWS,
+	}: FetchDataParams): Promise<FetchDataResult> => {
+		// Shared x-request-id ties both upstream calls together for tracing.
+		const requestId = generateRequestId();
+		const headers = buildHeaders(token, undefined, undefined, { requestId });
+
+		// Step 1: resolve the object's type — it decides the data endpoint.
+		const metaData = await postJson(
```
Rifdhan: Why do we need to do /metadata/search again? I thought we already did this in search_objects tool and returned all the necessary details to directly call the relevant /data API here.

saharsh-ts: Added optional object_type — pass the type from search_objects and we skip the lookup (one call). Kept optional so a bare GUID still resolves.

src/thoughtspot/fetch-data/fetch-data.ts
saharsh-ts: Done — enum ObjectType { Answer, Liveboard }.

src/servers/tool-definitions.ts
saharsh-ts: Removed — agent can count data_rows. Kept total_row_count (upstream total, not derivable when capped).

src/servers/tool-definitions.ts
saharsh-ts: Removed request_id. Still sent as x-request-id upstream for tracing.

src/servers/tool-definitions.ts
saharsh-ts: Trimmed output to just data. Removed id/name/type/description — all already known from search_objects.

src/servers/tool-definitions.ts
saharsh-ts: updated tool definitions.

src/servers/tool-definitions.ts
```diff
@@ -560,6 +630,8 @@ export const toolDefinitionsV2 = [
...
 			"",
+			"CRITICAL — explaining an object requires its DATA, never this metadata: the fields here (title, description, `query` tokens, tags) describe an object; they are NOT its contents. If the user asks to explain, describe, summarize, analyze, interpret, or say what any returned object contains, shows, means, or 'is about', you MUST call `fetch_data` for that object FIRST and ground your answer in the rows it returns. Do NOT answer such a request from this response alone, do NOT restate the metadata as an explanation, and do NOT merely offer to fetch — just call `fetch_data`. Use the `fetch_data` tool to do this — its own schema documents how to pass the object, so don't work the arguments out here. (Worksheets are the exception — they have no saved result to fetch.) And never distinguish or explain objects from their `description` or `query` tokens: those are authoring metadata, frequently stale and often copied verbatim across sibling vizzes, so two objects can share a byte-identical description while their data differs.",
```
Rifdhan: Claude app store guidelines prevent us from using terminology like "critical" or "must", as they do not want external apps dictating what the agent must do. Please soften the language to "should" or "can".

src/servers/tool-definitions.ts
```diff
@@ -580,6 +652,22 @@ export const toolDefinitionsV2 = [
...
+			"CALL THIS WHENEVER the user wants to explain, describe, summarize, analyze, interpret, or ask what a specific object contains or shows: any explanation of what an object contains must be grounded in the data this tool returns, so fetch it first with the object's `id`, then answer from the returned data.",
+			"Runs the object's saved question (its existing filters and columns, unchanged) against the current data and returns the result, so the rows reflect the latest data and may differ from when the object was first saved. The result is shaped to the object's type: an Answer returns a single tabular result, a Liveboard returns one tabular result per visualization (each with its visualization id and name). To pull a single visualization pinned on a Liveboard, pass the Liveboard GUID as `object_id` and the visualization GUID in `visualization_ids`. Each result includes the column names and data rows. Use the optional `max_rows` to bound the rows returned per visualization (defaults to 25).",
```
Rifdhan, line 669: Ditto here

src/thoughtspot/fetch-data/fetch-data.ts
```diff
@@ -0,0 +1,162 @@
...
+		objectType: knownType,
+		vizIds,
+		maxRows = FETCH_DATA_DEFAULT_MAX_ROWS,
+	}: FetchDataParams): Promise<FetchDataResult> => {
+		// x-request-id ties the upstream call(s) together for tracing.
+		const requestId = generateRequestId();
+		const headers = buildHeaders(token, undefined, undefined, { requestId });
+
+		// Resolve the object's type (decides the data endpoint) only when the
+		// caller didn't already supply it from a prior search_objects result.
```
Rifdhan: I think this is unnecessary additional complexity and would prefer to remove this extra logic. The caller MUST give us the type or we can just fail the request. I don't see any reason why the caller cannot give us the type.

src/servers/tool-definitions.ts
saharsh-ts: updated tool definitions

src/servers/tool-definitions.ts
saharsh-ts: updated

src/servers/tool-definitions.ts
saharsh-ts: this has been updated.

src/servers/tool-definitions.ts
saharsh-ts: this has been updated.

src/thoughtspot/fetch-data/fetch-data.ts
saharsh-ts: done

src/thoughtspot/get-object-data/get-object-data.ts
```diff
@@ -0,0 +1,146 @@
...
+
+// "Unbounded" record_size for Liveboards (they 500 if it can't hold the whole
+// viz). Max 32-bit signed int — the endpoint reads it as a GraphQL Int.
+export const LIVEBOARD_RECORD_SIZE = 2_147_483_647;
+
+// Only Answers and Liveboards expose fetchable data.
+enum ObjectType {
+	Answer = "ANSWER",
+	Liveboard = "LIVEBOARD",
+}
```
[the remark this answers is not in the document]
saharsh-ts: Done — added GET_OBJECT_DATA_SUPPORTED_TYPES; now the single source for both the Zod enum and the endpoint switch (dropped the local enum).

src/servers/mcp-server.ts
```diff
@@ -1101,6 +1111,44 @@ Provide this url to the user as a link to view the liveboard in ThoughtSpot.`;
...
+
+			return this.createStructuredContentSuccessResponse(
+				result,
+				`Fetched data for ${object_id} (${result.data.length} result(s))`,
+			);
+		} catch (error) {
+			// Surface the upstream message (e.g. status 401/500) so the failure is
+			// actionable rather than a generic "check the object id".
+			return this.createErrorResponse(
+				`Failed to fetch object data: ${(error as Error).message}`,
```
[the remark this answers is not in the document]
saharsh-ts: Done — guarded with error instanceof Error ? error.message : String(error).

src/thoughtspot/get-data/get-data.ts
```diff
@@ -0,0 +1,144 @@
...
+	GetObjectDataViz,
+} from "./get-object-data-types";
+
+// Default row cap per visualization; unbounded results overwhelm LLM context.
+// Keep in sync with the `max_rows` description in tool-definitions.ts.
+export const GET_OBJECT_DATA_DEFAULT_MAX_ROWS = 25;
+
+// "Unbounded" record_size for Liveboards (they 500 if it can't hold the whole
+// viz). Max 32-bit signed int — the endpoint reads it as a GraphQL Int.
+export const LIVEBOARD_RECORD_SIZE = 2_147_483_647;
```
mouryabalabhadra, line 16: Liveboard fetches send record_size = 2,147,483,647 and apply max_rows only client-side, so the Worker downloads the entire result set of every visualization before slicing to 25 rows.

src/servers/tool-definitions.ts
```diff
@@ -119,13 +120,13 @@ const SearchObjectResultSchema = z.object({
...
 		.describe(
-			"The GUID to pass to fetch_data as `object_id`. For a visualization pinned on a Liveboard this is the parent Liveboard's GUID (see `visualization_id`).",
+			"The GUID to pass to get_object_data as `object_id`. For a visualization pinned on a Liveboard this is the parent Liveboard's GUID (see `visualization_id`).",
 		),
 	visualization_id: z
 		.string()
 		.optional()
 		.describe(
-			"Set only when this result is a specific visualization on a Liveboard: `object_id` is the Liveboard and this is the visualization. Pass it to fetch_data as `visualization_ids` to fetch just this viz.",
+			"Set only when this result is a specific visualization on a Liveboard: `object_id` is the Liveboard and this is the visualization. Pass it to get_object_data as `visualization_ids` to fetch just this viz.",
```
mouryabalabhadra: we are unconditionally instructing the model to call get_object_data but it sbehind privilege

src/servers/mcp-server.ts
```diff
@@ -1141,6 +1151,44 @@ Provide this url to the user as a link to view the liveboard in ThoughtSpot.`;
...
+				vizIds: visualization_ids,
+				maxRows: max_rows,
+			});
+
+			return this.createStructuredContentSuccessResponse(
+				result,
+				`Fetched data for ${object_id} (${result.data.length} result(s))`,
+			);
+		} catch (error) {
+			// Surface the upstream message (e.g. status 401/500) so the failure is
```
mouryabalabhadra: this is too generic catch all error.  401 errors have the dispatch-layer 401 handler to return explicit reauthentication message

src/servers/mcp-server.ts
```diff
@@ -1141,6 +1151,44 @@ Provide this url to the user as a link to view the liveboard in ThoughtSpot.`;
 		);
 	}
 
+	@WithSpan("call-get-object-data")
+	async callGetObjectData(
```
mouryabalabhadra: callGetObjectData checks canDownloadData() but doesn't wait for ensureSessionInfo() 

it should do similar to listTools

src/servers/tool-definitions.ts
saharsh-ts: Fixed

src/servers/mcp-server.ts
saharsh-ts: done

src/servers/mcp-server.ts
saharsh-ts: added `await this.ensureSessionInfo()` before the gate, same as `listTools`

src/thoughtspot/get-data/get-data.ts
saharsh-ts: ack. The Liveboard /data endpoint 500s when record_size is below the viz's full row count, so it can't be bounded server-side — `record_size` is int32-max and `max_rows` caps client-side.

so we fetch complete liveboard vizzes, and then use min(numViz, 25) objects to explain the liveboard.

mouryabalabhadra:
@copilot review the pr

src/servers/mcp-server.ts
```diff
@@ -417,6 +419,24 @@ export class MCPServer extends BaseMCPServer {
 			);
 		}
 
+		// Hide get_object_data if the user lacks the data-download privilege, and
+		// drop search_objects' directive to call it (else it points at a hidden tool).
+		if (!this.canDownloadData()) {
```
mouryabalabhadra, line 432: canDownloadData() requires sessionInfo.privileges. If user switches org, stale cache produces wrong answer.

@rohitthughtspot @saharsh-ts 

can we update the logic in callSwitchOrg, this.sessionInfo = undefined after setActiveOrg, then ensureSessionInfo() refetches with org token on next gated call. Refetch after ensureActiveOrg in postInit also needed for reconnect with persisted active org.

src/servers/mcp-server.ts
```diff
@@ -1141,6 +1165,52 @@ Provide this url to the user as a link to view the liveboard in ThoughtSpot.`;
...
+		// Enforce the gate even if a client calls the hidden tool directly.
+		if (!this.canDownloadData()) {
+			return this.createErrorResponse(
+				"You do not have permission to download data (requires the data-download privilege).",
+				"get_object_data forbidden",
+			);
+		}
+
+		const { object_id, object_type, visualization_ids, max_rows } =
+			GetObjectDataInputSchema.parse(request.params.arguments);
```
mouryabalabhadra: every tool handler runs Schema.parse outside its try.

object_type enum ["ANSWER","LIVEBOARD"] and search_objects type is LIVEBOARD | LIVEBOARD_VIZ | ANSWER | WORKSHEET.

first cross-tool handoff where one tool's output value is rejected by next tool's schema

src/servers/mcp-server.ts
```diff
@@ -417,6 +419,24 @@ export class MCPServer extends BaseMCPServer {
...
+		// Hide get_object_data if the user lacks the data-download privilege, and
+		// drop search_objects' directive to call it (else it points at a hidden tool).
+		if (!this.canDownloadData()) {
+			tools = tools
+				.filter((tool) => tool.name !== ToolName.GetObjectData)
+				.map((tool) =>
+					tool.name === ToolName.SearchObjects
+						? {
+								...tool,
+								description: tool.description.replace(
```
mouryabalabhadra: Directive stripped by exact-substring replace

src/servers/mcp-server.ts
saharsh-ts: Fixed — switch_org clears sessionInfo so the gate refetches under the new org token; postInit refetches after the org mint on reconnect.

src/servers/mcp-server.ts
saharsh-ts: Fixed — object_type now accepts LIVEBOARD_VIZ (routes via the parent Liveboard endpoint)

src/servers/mcp-server.ts
saharsh-ts: Fixed — strip now line-matches the directive (split/filter) instead of an exact \n\n substring, so it's position-independent.

src/servers/mcp-server.ts
```diff
@@ -417,6 +430,30 @@ export class MCPServer extends BaseMCPServer {
...
+		if (!this.canDownloadData()) {
+			tools = tools
+				.filter((tool) => tool.name !== ToolName.GetData)
+				.map((tool) =>
+					tool.name === ToolName.SearchObjects
+						? {
+								...tool,
+								// Drop the directive line (its own paragraph), then collapse the
+								// blank-line gap. Line-match, not substring, so it's position-safe.
+								description: tool.description
```
Rifdhan: This seems pretty hacky, let's just update the description of search_objects to not mention get_data at all. I don't think it's really necessary anyway, the agent can understand when to use this additional tool.

saharsh-ts: Done — removed the strip and dropped the `get_data` mention from `search_objects`. 
Kept just a generic "explanations should come from the data, not the metadata" note.

src/thoughtspot/get-data/get-data.ts
```diff
@@ -0,0 +1,170 @@
...
+	const named = content.column_names ?? [];
+	const rowKeys = new Set<string>();
+	for (const row of rawRows) {
+		if (isObjectRow(row)) {
+			for (const k of Object.keys(row)) {
+				rowKeys.add(k);
+			}
+		}
+	}
+	const columns = [
```
mouryabalabhadra: this  drops any column that no row carries as a key. is that intentional?

src/thoughtspot/get-data/get-data-types.ts
```diff
@@ -0,0 +1,31 @@
+// Interfaces for the `get_data` tool: full data of an Answer/Liveboard GUID.
+
+export interface GetDataParams {
+	// GUID of an Answer or Liveboard, typically from `search_objects`.
+	objectId: string;
+	// Required type ("ANSWER"/"LIVEBOARD") from search_objects; picks the endpoint.
+	objectType: string;
```
mouryabalabhadra: shouldn't this be "ANSWER"|"LIVEBOARD"

src/servers/mcp-server.ts
mouryabalabhadra: sessionInfo privileges are not refreshed when another session runs switch_org, so the get_data gate can use the wrong org's privileges

src/thoughtspot/get-data/get-data.ts
saharsh-ts: Fixed — now keeps every declared column even if no row carries it; missing cells read as null.

src/thoughtspot/get-data/get-data-types.ts
saharsh-ts: Done — typed as the supported-types union, not a bare string.

src/servers/mcp-server.ts
saharsh-ts: `canDownloadData()` and data token are both from the connection's own in-memory org, so they never mismatch. 

Do you want a live connection to immediately follow another connection's org switch (converge all connections)? That needs a per-call store read + re-mint. 
OR 
is per-connection-until-reconnect fine?

PR #184, opened 2026-07-12
Repair session info on demand + fix callTool token reconcile

rohitthughtspot:
## Problem
After a cold-start reconnect (~24h, once the frozen props access token has expired), `list_orgs`/`switch_org` disappear and datasource-discovery gating breaks — even though the keep-warm token is valid.

## Cause
`sessionInfo` is fetched once at init via `getSessionInfo()`, which runs **before** `postInit` reconciles the keep-warm global token from the DO. On that reconnect the fetch authenticates with the dead frozen props token, fails, and leaves `sessionInfo` null for the instance lifetime. Org-tool visibility and datasource gating read `sessionInfo`, so they mis-gate — while the DO still holds a valid token (data tool calls, which reconcile it at call time, keep working).

## Fix
- **`ensureSessionInfo()`** — refetch session info when it's null, called at the top of `listTools()`/`callTool()`. By then `postInit` has loaded the kept-warm token, so the refetch authenticates with it. No-op (no network call) once populated — happy path unaffected.
- **`callTool` per-call reconcile keyed on `!this.activeOrgToken`** — so a data call always gets a freshly reconciled global token when no org token drives it (covers non-orgs clusters and the no-active-org case).

## Test
Regression test reproduces the failure (init `getSessionInfo` fails on an expired props token; DO holds a valid token) and asserts the repair restores `sessionInfo` and the org tools. Verified it fails without the fix. Server suites pass.

### DIFF
--- a/src/servers/mcp-server-base.ts
+++ b/src/servers/mcp-server-base.ts
@@ -61,6 +61,8 @@ export interface Context {
 export abstract class BaseMCPServer extends Server {
 	protected trackers: Trackers = new Trackers();
 	protected sessionInfo: any;
+	// In-flight ensureSessionInfo() refetch, so concurrent callers share one fetch.
+	private sessionInfoPromise?: Promise<void>;
 
 	constructor(
 		protected ctx: Context,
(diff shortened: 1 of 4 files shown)


mouryabalabhadra:
@copilot review pr

Rifdhan (approved):

PR #192, opened 2026-07-30
Provide safe defaults for flags if sessionInfo is unavailable

Rifdhan:
- Temporary fix to ensure flags are passed correctly in most cases
- Update tests

### DIFF
--- a/src/servers/mcp-server-base.ts
+++ b/src/servers/mcp-server-base.ts
@@ -89,15 +89,21 @@ export abstract class BaseMCPServer extends Server {
 	 * Whether Spotter data source discovery (Auto Mode) is enabled
 	 */
 	protected isSpotterDataSourceDiscoveryEnabled(): boolean {
-		return this.sessionInfo?.isSpotterDataSourceDiscoveryEnabled === true;
+		if (!this.sessionInfo) {
+			console.warn(
+				"Session info not available when checking data source discovery flag",
+			);
+			return true;
+		}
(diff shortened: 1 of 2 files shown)


src/servers/mcp-server-base.ts
```diff
@@ -89,15 +89,21 @@ export abstract class BaseMCPServer extends Server {
 	 * Whether Spotter data source discovery (Auto Mode) is enabled
 	 */
 	protected isSpotterDataSourceDiscoveryEnabled(): boolean {
-		return this.sessionInfo?.isSpotterDataSourceDiscoveryEnabled === true;
+		if (!this.sessionInfo) {
+			console.warn(
+				"Session info not available when checking data source discovery flag",
+			);
+			return true;
+		}
```
shikharbsar, line 97: should we return true here or false?

src/servers/mcp-server-base.ts
```diff
@@ -107,7 +113,13 @@ export abstract class BaseMCPServer extends Server {
 	 * Whether Spotter chat history (save chat) is enabled
 	 */
 	protected isSpotterChatHistoryEnabled(): boolean {
-		return this.sessionInfo?.isSpotterChatHistoryEnabled === true;
+		if (!this.sessionInfo) {
+			console.warn(
+				"Session info not available when checking chat history flag",
+			);
+			return true;
+		}
```
shikharbsar, line 121: same as above

shikharbsar (approved):

PR #197, opened 2026-08-01
SCAL-320648 mcp changes to support spotter model through tools

pradeepburugula:
## Summary
Adds a **Spotter Model** tool surface that lets an MCP client build (or edit) a ThoughtSpot data model conversationally through the Lumos agentic backend — the agent plans, auto-builds tables/joins/columns/formulas, and saves the result as a worksheet.

The flow deliberately mirrors the existing Spotter tools: a three-step session (create → send → get updates) plus a terminal save, the same shape as `create_analysis_session` / `send_session_message` / `get_session_updates` + `create_dashboard`.

**Diff size:** 24 files against `main`, split **1,700 lines of production code / 2,057 lines of tests**. Production code grew only ~200 lines net over the previous revision of this branch (the fourth tool, the privilege gate, and per-file module boilerplate, less the duplicated storage plumbing that was deleted) — the rest of the growth is the test suite, which took the branch from failing CI's 85% coverage gate at 78.4% to passing at 92.1%.

## Tools (4)

| Tool | Purpose |
|---|---|
| `create_model_session` | Opens a session. `connection_identifier` to build a NEW model, or `model_identifier` to open an existing one for editing. Returns a `model_session_id` used by the other three. Builds nothing on its own. |
| `send_model_message` | Sends one instruction (a model description, a correction, or `selected_option_ids` to answer a clarification). Returns as soon as it is accepted. |
| `get_model_updates` | Drains the builder's updates for the instruction in flight. Call until `is_done` is true. |
| `finalize_model` | Review (no `confirm`) → save (`confirm=true`); returns the model URL. |

## Behavior & guardrails
- **New model vs. edit existing:** `connection_identifier` (new) or `model_identifier` (open a saved model; edits and the save go back to it). An edit-save with no `name` preserves the existing name/description.
- **Approval gates (soft, description-enforced):** on an `mrd` update (the plan), the assistant must stop and get the user's explicit approval before building — never auto-approve; and it must confirm the specific formulas with the user before adding them.
- **Connection safety:** starting a session with neither a connection nor a model is rejected, and the schema tells the assistant to ask the user and confirm the exact GUID (never guess/reuse/assume). The server cannot detect an *invented* GUID — only the missing case — so this remains guidance plus one hard gate.
- **One instruction at a time:** sending a second instruction while a turn is still streaming is **rejected** with a "still working on the previous instruction" error rather than interleaving two turns' updates into one stream. This falls out of reusing the analytical-session per-turn `initializeConversation`.
- **Permission-gated:** see *Permissions* below — users who cannot create data models are not offered these tools at all.

## Streaming & output (no chain-of-thought leak)
- Raw `text_chunk` is **not** streamed. The SSE consumer accumulates silently and emits ONE cleaned update per turn at `MESSAGE_END` — an `mrd` (plan, scaffolding-stripped) on planning turns, else a cleaned `text`.
- Live progress flows via `todo` (task tracker: Tables → Joins → Columns), `notification`, `model_state` and `choice` updates; `get_model_updates`' description instructs the client to render the `todo` tracker and notifications as progress.
- **Poll window:** `get_model_updates` long-polls for ~60s (120 × 500 ms), deliberately longer than the analytical-session tool's ~10s. A model build runs 1–2 minutes and each MCP call carries ~10s of connector overhead, so a full build costs 2–3 calls instead of ~12. Same tool shape as Spotter, different tuning.
- **Finalize summary is structured, not prose:** the review step reads the actual model via `fetchWorksheetModel` and renders "The model contains: N tables (names), M joins, K columns" — no model-generated text, so no scaffolding leak and no empty summary.
- Polling before any instruction has been sent returns a clear "no updates yet" error rather than a 500.

## Storage — reuses the Spotter conversation store
No parallel storage mechanism. Model updates ride the **same** append/read-bookmark message stream the analytical-session tools use, isolated by the conversation id the Durable Object is routed by:
- `send_model_message` calls the same per-turn `initializeConversation`, which is also what clears the previous turn's done flag and provides the concurrent-instruction rejection above.
- The background stream consumer appends through `appendMessages`; `get_model_updates` drains with `getNewMessages`.
- The one genuinely new piece of state is the model session's scalar data (bach transaction id, generation number, generation working set, session cookie, pending clarification), which has no analytical-session equivalent. It lives in a **generic, opaque `POST/GET /state` blob route** — usable by either flow, and the DO does not interpret the payload.
- The message reader now **skips the read-bookmark write on an idle poll**, which also benefits the analytical-session poll.
- **No new DO class and no migration** — additive routes on the existing `ConversationStorageServerSQLite`.

## Upstream calls
- **Init:** `POST /lumos/api/v2/conversation/model/init` — the model-specific entry point (this replaces the earlier `/chat/init` call). Lumos opens the Bach transaction itself. The request is strict protobuf, so the wire keys are the proto field names — `connectionGuid` for a new model, `modelGuid` to open an existing one — and only the one that applies is sent (`modelGuid` wins upstream). An unrecognised key is silently dropped rather than rejected, surfacing as a misleading "identifier is required" 400, so the exact names matter. The response is parsed tolerantly (snake_case or camelCase).
- **Messaging:** `POST /lumos/api/v2/conversation/chat/{conversation_id}/stream`, mirroring the UI body (camelCase keys, `pinnedClientGenerationNumber`, explicit `choice`, `isRestoreRequest`), with a minted session cookie forwarded for cookie-only backend tools (FormulaGen).
- **Save/read:** the bach worksheet-editor op (`/prism/?op=WorksheetOperation`) — `SAVE_WORKSHEET_REQUEST` for the save (needs the full `genNoWorkingSet`, since one build spans many generations) and `FETCH_WORKSHEET_MODEL_REQUEST` for the finalize summary.

## Code layout
The feature lives in its own directory, following the `search-objects/` convention rather than extending `thoughtspot-client.ts`:

| File | Contents |
|---|---|
| `spotter-model/spotter-model-client.ts` | `addSpotterModel()` registrar: `model/init`, `chat/{id}/stream`, `saveModel`, `fetchWorksheetModel`, `mintSessionCookie` |
| `spotter-model/spotter-model-stream.ts` | SSE consumer + `advanceGeneration` / `looksLikeMrd` / `extractMrdPlan` / `cleanTurnText` |
| `spotter-model/spotter-model-mapper.ts` | `buildChoiceAnswer`, `summarizeWorksheetModel` |
| `spotter-model/spotter-model-types.ts` | `ModelSessionState`, `ModelUpdate`, `ModelStreamSink`, upstream request/response shapes |
| `spotter-model/spotter-model-queries.ts` | bach GraphQL operations |
| `spotter-model/spotter-model-constants.ts` | endpoints, poll window, MRD markers |

The stream consumer writes through a `ModelStreamSink` interface rather than taking the storage client, so the `thoughtspot/` layer does not depend on the Durable Object client and the consumer is unit-testable with a fake.

## Permissions
All four tools are gated on data-modeling privileges — `DATAMANAGEMENT` ("Can manage data"), `CAN_MANAGE_WORKSHEET_VIEWS_TABLES` (the granular RBAC "Can manage data models") or `ADMINISTRATION`:
- `canManageDataModels()` in `mcp-server-base.ts` reads the already-fetched `sessionInfo.privileges` — **no extra API call** — and **fails closed**, unlike the neighbouring feature-flag helpers which default to enabled.
- `listTools` filters all four out, so an unprivileged user's toolspace is not polluted with tools they cannot use.
- `dispatchTool` rejects a call to any of the four as well. This is not just defence in depth: client tool lists are **cached**, so a call can arrive after a privilege change or org switch. It repairs session info first, so a transient fetch failure does not deny a user who does have the privilege.

## Reviewer guide

| # | File | What to look at |
|---|---|---|
| 1 | `src/servers/tool-definitions.ts` | The 4-tool contract: schemas, per-tool descriptions, `ModelUpdate` shape, `MODEL_TOOL_NAMES` |
| 2 | `src/servers/mcp-server.ts` | `callCreateModelSession` / `callSendModelMessage` / `callGetModelUpdates` / `callFinalizeModel`, `pollModelUpdates`, `modelStreamSink`, and the two privilege gates (`listTools`, `dispatchTool`) |
| 3 | `src/thoughtspot/spotter-model/` | The module above — client, stream consumer, mappers, types |
| 4 | `src/servers/conversation-storage-server.ts` / `storage-service.ts` | The generic `/state` route and the idle-poll write skip; note what was *removed* |
| 5 | `src/servers/mcp-server-base.ts` | `canManageDataModels()` and the privilege list |
| 6 | `src/servers/version-registry.ts` | V3 registered as `latest` |

## Testing
- **815 tests passing**; `biome lint` and `biome check` clean; no new `tsc` errors beyond the repo baseline.
- Coverage **92.1% statements / 87.5% branches / 92.4% functions / 92.1% lines**, above the 85% gate and at or above `main`'s baseline. (The earlier revision of this branch was at 78.4%/72.1% and failed CI's coverage gate.)
- New tests: unit coverage for the SSE consumer, the mappers and every upstream call; plus an integration spec driving all four tools through a **real** `ConversationStorageServerSQLite` (per-turn initialize, background consumer write-through, long-poll drain, concurrent-instruction rejection, choice round-trip, generation working-set accumulation); plus privilege-gate tests for each accepted privilege, the unprivileged case and the fail-closed case.
- Verified live on a dev cluster after deploy: create → send → poll → finalize.

### DIFF
--- a/src/servers/conversation-storage-server.ts
+++ b/src/servers/conversation-storage-server.ts
@@ -1,6 +1,9 @@
 import { isBoolean } from "lodash";
 import type {
 	Message,
+	ModelSessionState,
+	ModelUpdate,
+	ModelUpdatesState,
 	RawMessage,
 	StreamingMessagesState,
 } from "../thoughtspot/types";
@@ -12,6 +15,15 @@ const MESSAGE_KEY_PREFIX = "message-";
 const IS_DONE_KEY = "is-done";
 const WRITE_BOOKMARK_KEY = "write-bookmark";
 const READ_BOOKMARK_KEY = "read-bookmark";
+
+// Spotter model-session state: single blob in a separate DO instance (routed by model_session_id).
+const MODEL_SESSION_KEY = "model-session";
+
+// Streamed model updates: own write/read bookmarks + per-index keys; consumer appends while reader polls (each DO op is atomic).
+const MODEL_UPDATE_KEY_PREFIX = "model-update-";
+const MODEL_WRITE_BOOKMARK_KEY = "model-write-bookmark";
+const MODEL_READ_BOOKMARK_KEY = "model-read-bookmark";
+const MODEL_UPDATES_DONE_KEY = "model-updates-done";
 
 /**
  * A Durable Object that stores streaming conversation messages and exposes them over HTTP.
@@ -57,6 +69,36 @@ export class ConversationStorageServerSQLite {
 
 				case "GET /messages": {
 					const state = await this.getNewMessagesAndUpdateBookmark();
+					return Response.json(state);
+				}
+
+				case "POST /model-put": {
+					const body = (await request.json()) as ModelSessionState;
+					await this.putModelSession(body);
+					return Response.json({ ok: true });
+				}
+
+				case "GET /model-get": {
+					const state = await this.getModelSession();
+					return Response.json(state);
+				}
+
+				case "POST /model-append": {
+					const body = (await request.json()) as {
+						updates?: ModelUpdate[];
+						isDone?: boolean;
+						resetDone?: boolean;
+					};
+					await this.appendModelUpdates(
+						body.updates ?? [],
+						body.isDone ?? false,
+						body.resetDone ?? false,
+					);
+					return Response.json({ ok: true });
+				}
+
+				case "GET /model-updates": {
+					const state = await this.getNewModelUpdatesAndAdvanceBookmark();
 					return Response.json(state);
 				}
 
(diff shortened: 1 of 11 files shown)


rohitthughtspot (dismissed):

src/servers/version-registry.ts
```diff
@@ -60,7 +64,13 @@ export const VERSION_REGISTRY: VersionConfig[] = [
 		description: "Spotter3 agent conversation tools released",
 	},
 	{
-		version: ["latest", "2026-05-01"],
+		version: ["latest", "2026-07-31"],
```
Rifdhan, line 67: Let's update this to Sep 25 to give us time to ship after hackathon

src/servers/conversation-storage-server.ts
```diff
@@ -142,33 +160,50 @@ export class ConversationStorageServerSQLite {
...
+					console.warn(
+						`Expected message at index ${i} for conversation ${this.conversationId} not found`,
+						{ readBookmark, writeBookmark },
+					);
+					continue;
+				}
+				newMessages.push(message);
+			}
+
+			await this.state.storage.put<number>(READ_BOOKMARK_KEY, writeBookmark);
```
Rifdhan, line 185: Could you explain the changes being made to getNewMessagesAndUpdateBookmark here? Do we need these?

src/servers/mcp-server.ts
```diff
@@ -1162,4 +1219,381 @@ Provide this url to the user as a link to view the liveboard in ThoughtSpot.`;
...
+			if (isDone) {
+				break;
+			}
+			await new Promise((resolve) =>
+				setTimeout(resolve, MODEL_POLL_INTERVAL_MS),
+			);
+		}
+		return { updates, isDone };
+	}
+	@WithSpan("call-finalize-model")
```
Rifdhan, line 1519: Nit: leave a blank line between function defs

src/servers/tool-definitions.ts
```diff
@@ -475,8 +475,211 @@ export enum ToolName {
...
+];
+
+// ── Spotter Model (V3) schemas ─────────────────────────────────────────────
+export const CreateModelSessionInputSchema = z.object({
+	connection_identifier: z
+		.string()
+		.optional()
+		.describe(
+			"GUID of the data-warehouse connection to build a NEW model on (e.g. " +
+				"'xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx'). You MUST ask the user which connection to " +
```
Rifdhan, line 503: We cannot use "MUST" language in tool descriptions or we will violate Claude app store rules. Change all phrasing to "should".

pradeepburugula: addressed the comment

src/servers/mcp-server.ts
pradeepburugula: changes done

src/servers/conversation-storage-server.ts
pradeepburugula: **What changed.** Just a guard. The existing body is now wrapped in `if (writeBookmark > readBookmark)`, so an idle poll — nothing written since the last read — returns early instead of falling through it.                                                                                                                                        
                                                                                                                                                                              
**Why it's behaviourally a no-op.** When the bookmarks are equal, the old path already did nothing useful:                                                                  
                                                                                                                                                                              
`keys` was empty, and `getInBatches([])` loops zero times, so there was never a read to skip;                                                                             
the message loop body never executed;                                                                                                                                     
`put(READ_BOOKMARK_KEY, writeBookmark)` stored the value the bookmark already held.                                                                                       
                                                                                                                                                                              
Both branches return the same `{ messages: [], isDone }`, so no caller can tell the difference.                                                                             
                                                                                                                                                                              
**Do we need it.** Not for correctness — the model tools work fine without it. What it buys is write volume. `get_model_updates` long-polls at 500ms for up to 120 iterations (~60s per call) while a build runs 1–2 minutes, so before the guard every call issued up to ~120 Durable Object writes that changed nothing, each one taking the DO write path and contending with the stream consumer appending to the same object. `get_session_updates` polls the same way, so V2 gets the same saving — that's the only reason this lands in shared code instead of staying on the V3 path.

PR #199, opened 2026-08-05
[SCAL-327336] Fix org tools disappearing after token expiry: reconcile token before getSessionInfo

rohitthughtspot:
[SCAL-327336]

## Problem
After a cold-start reconnect (~24h, once the frozen props access token has expired), `list_orgs`/`switch_org` disappear and feature-flag gating misbehaves — even though the keep-warm token is still valid.

## Cause
`sessionInfo` is fetched once at init via `getSessionInfo()`, which previously ran **before** `postInit` reconciled the keep-warm global token from the token-store DO. On a post-expiry reconnect the init fetch authenticated with the dead frozen props token, failed, and left `sessionInfo` null for the DO's lifetime. Org-tool visibility and flag gating read `sessionInfo`, so they silently degraded — while the DO still held a valid token (data calls, which reconcile the DO token at call time, kept working).

## Fix
- **Reorder** — a `preInit()` hook reconciles the global token from the token-store DO **before** `initializeService()`/`getSessionInfo`, so the init fetch uses the kept-warm token and `sessionInfo` populates correctly.
- **`ensureSessionInfo()`** — fallback in `listTools` that refetches session info if it's still null (e.g. a transient init failure), guarded by an in-flight promise so concurrent list calls share one fetch (no duplicate `getSessionInfo` / MixpanelTracker).
- **`isOrgsEnabled()`** defaults to true when `sessionInfo` is absent, so org tools stay visible in the brief window before the refetch completes.
- **`callTool`** reconciles the global token per call keyed on the absence of an active org token.
- Extract `postInit`'s active-org bootstrap into `ensureActiveOrg()`.

## Test
Regression test reproduces the failure (init `getSessionInfo` fails on an expired props token; DO holds a valid token) and asserts the repair restores `sessionInfo` and the org tools. Full suite passes (695 tests).

Supersedes #184 (rebased onto latest main as a single clean commit).

[SCAL-327336]: https://thoughtspot.atlassian.net/browse/SCAL-327336?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/servers/mcp-server-base.ts
+++ b/src/servers/mcp-server-base.ts
@@ -112,6 +112,19 @@ export abstract class BaseMCPServer extends Server {
 	}
 
 	/**
+	 * Whether Spotter chat history (save chat) is enabled
+	 */
+	protected isSpotterChatHistoryEnabled(): boolean {
+		if (!this.sessionInfo) {
+			console.warn(
+				"Session info not available when checking chat history flag",
+			);
+			return true;
+		}
+		return this.sessionInfo.isSpotterChatHistoryEnabled === true;
+	}
+
+	/**
 	 * Fallback: refetch session info if it's still null (preInit normally loads the
 	 * token so the init-time fetch succeeds, but if that failed too — e.g. a
 	 * transient error — repair on demand here). Called by callers that read
@@ -128,19 +141,6 @@ export abstract class BaseMCPServer extends Server {
 			});
 		}
 		await this.sessionInfoPromise;
-	}
-
-	/**
-	 * Whether Spotter chat history (save chat) is enabled
-	 */
-	protected isSpotterChatHistoryEnabled(): boolean {
-		if (!this.sessionInfo) {
-			console.warn(
-				"Session info not available when checking chat history flag",
-			);
-			return true;
-		}
-		return this.sessionInfo.isSpotterChatHistoryEnabled === true;
 	}
 
 	/**
--- a/src/servers/mcp-server-base.ts
+++ b/src/servers/mcp-server-base.ts
@@ -88,12 +88,21 @@ export abstract class BaseMCPServer extends Server {
 	}
 
 	/**
-	 * Whether Spotter data source discovery (Auto Mode) is enabled
+	 * Whether Spotter data source discovery (Auto Mode) is enabled.
+	 *
+	 * Fails OPEN (returns true) when sessionInfo is unavailable. sessionInfo can
+	 * be null if getSessionInfo failed at init (e.g. a transient cluster error, or
+	 * historically a post-expiry reconnect where the fetch used the dead frozen
+	 * props token — now fixed by the preInit token reorder). This flag only tunes
+	 * how a tool runs (auto datasource discovery), not whether it may run, so
+	 * defaulting it on during a brief session-info gap is the least-surprising,
+	 * non-breaking choice. Contrast getActiveToken(), which fails CLOSED because a
+	 * wrong answer there would send a dead token upstream.
 	 */
 	protected isSpotterDataSourceDiscoveryEnabled(): boolean {
 		if (!this.sessionInfo) {
 			console.warn(
-				"Session info not available when checking data source discovery flag",
+				"Session info not available when checking data source discovery flag; defaulting ON (fail-open)",
 			);
 			return true;
 		}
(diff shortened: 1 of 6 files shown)


Rifdhan:
@copilot review this change and identify any gaps or potential issues, but don't make any code changes yourself

rohitthughtspot:
Thanks for the thorough review — dispositions on the raised points:

**1. `ensureSessionInfo` in `listTools` but not `callTool` — intentional, not adding it.**
`sessionInfo` at `callTool` has only two kinds of consumers, and neither needs a refetch there:
- **Tool availability** (`isOrgsEnabled()` → `areOrgToolsAvailable()`, gating `list_orgs`/`switch_org`). Availability is authoritatively decided in `listTools`; the re-check in `callTool` is a cheap guard against a client calling a tool it was never offered. With `sessionInfo` null it returns a clean "only available on orgs-enabled clusters" message — safe, and the tool was already hidden at list time.
- **Feature-behavior flags** `isSpotterDataSourceDiscoveryEnabled()` / `isSpotterChatHistoryEnabled()` passed into `createAgentConversation`. These are *how* the tool runs, not *whether*. They **fail-open (return true)** when `sessionInfo` is null.

Adding `ensureSessionInfo()` to `callTool` would spend a network round-trip to make two safe, fail-open flags marginally more accurate. Bad trade. With the new `preInit` ordering (`preInit` loads the kept-warm token → `initializeService()`/`getSessionInfo` runs with a valid token before any handler fires), `sessionInfo` is normally populated at init anyway; the `listTools` `ensureSessionInfo` is itself a belt-and-suspenders fallback for a transient init-fetch failure. Noting the fail-open behavior here explicitly for the record.

**2. Unbounded refetch during an outage — reconsidered, intentionally left as-is (no negative-cache).**
The refetch only fires when `sessionInfo` is null, which after the `preInit` reordering means `getSessionInfo` failed at init *even with a valid token* — i.e. the cluster itself is erroring. In that state it's **one refetch per tool call**, and tool calls are human-paced (a Claude session issues one every few seconds at most), so the rate is bounded by client request rate, not a loop. Concurrent callers are already coalesced by the shared in-flight promise. A negative-cache would only matter as retry-amplification against an already-down cluster at large fan-out — and even then these probes are marginal next to whatever is causing the outage, while the session degrades gracefully (org tools fail-closed, feature flags fail-open) rather than breaking. Not worth the added state/back-off logic for this workload's shape.

**3. `callTool` reconcile condition widened (`oauth && !activeOrgToken`) — intended, calling it out.** Correct that this now also reconciles the global token for org-enabled clusters with no active org selected (old `!areOrgToolsAvailable()` skipped that). That is the intended behavior — an org-capable user with no active org authenticates with the global token, so it should be reconciled per call — but it is a behavior change beyond the original description and affects all OAuth users on org clusters who haven't selected an org.

**4. `grantHasRefreshToken` set before the inner try — no change.** The only code before the assignment is the `authMode !== "oauth"` return and a `typeof` read; neither can throw, so the theoretical "silently disabled" path isn't reachable.

**5. Unused `recorder` on `ensureActiveOrg` — done, dropped the dead param.**

**6. Test asserting `trackers.size` — done, replaced with `warmSessionInfoCalls.count`,** which measures the coalescing behavior directly without coupling to `Trackers` internals.

src/servers/mcp-server-base.ts
```diff
@@ -85,37 +87,52 @@ export abstract class BaseMCPServer extends Server {
 		);
 	}
 
-	/**
-	 * Whether Spotter data source discovery (Auto Mode) is enabled
-	 */
+	// Fails open (true) when sessionInfo is unavailable: this tunes tool behavior,
+	// not availability, so defaulting on during a session-info gap is non-breaking.
```
Rifdhan: Please revert these changes to comments and logs in functions where the logic is not changing

src/servers/mcp-server-base.ts
```diff
@@ -85,37 +87,52 @@ export abstract class BaseMCPServer extends Server {
...
 		return this.sessionInfo.orgsEnabled === true;
 	}
 
-	/**
-	 * Whether Spotter chat history (save chat) is enabled
-	 */
+	// Refetch sessionInfo if the init fetch left it null (e.g. a transient error).
+	// preInit normally loads the token first so the init fetch succeeds; this is the
+	// fallback. No-op once populated; concurrent callers share one in-flight fetch.
+	protected async ensureSessionInfo(): Promise<void> {
```
Rifdhan, line 134: Move this after the block of flag check utils (after isSpotterChatHistoryEnabled)

src/servers/mcp-server-base.ts
rohitthughtspot: Ack

src/servers/mcp-server-base.ts
rohitthughtspot: Ack

Rifdhan (approved):

PR #200, opened 2026-08-06
Add raw Spotter output mode behind a flag

Rifdhan:
- Client can specify via URL flag if they want the full raw Spotter output instead of the simplified version we are providing
- Pass along the unmodified Spotter event stream if this flag has been set
- No impact when using v1 tools, this is only for v2 tools
- Update tests and add coverage

### DIFF
--- a/src/servers/tool-definitions.ts
+++ b/src/servers/tool-definitions.ts
@@ -537,10 +537,8 @@ export const toolDefinitionsV1 = [
 	},
 ];
 
-const GET_SESSION_UPDATES_BASE_TOOL_DEFINITION = {
+const GET_SESSION_UPDATES_SHARED_TOOL_DEFINITION = {
 	name: ToolName.GetSessionUpdates,
-	description:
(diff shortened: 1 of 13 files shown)


Rifdhan:
@copilot review this change. Make suggestions or identify any gaps but don't make any code changes yourself.

src/servers/tool-definitions.ts
```diff
@@ -589,18 +615,15 @@ export const toolDefinitionsV2 = [
...
 		outputSchema: z.toJSONSchema(GetSessionUpdatesOutputSchema),
-		annotations: {
-			title: "Get Analysis Session Updates",
-			readOnlyHint: true,
-			destructiveHint: false,
-			openWorldHint: false,
-		},
+	},
+	{
+		...GET_SESSION_UPDATES_BASE_TOOL_DEFINITION,
```
mouryabalabhadra: description is same for both

rohitthughtspot (approved):
LGTM

